PR #2, opened 2021-07-14
Deploy crd

Tal-or:
Deploy/Undeploy CRD

### DIFF
--- a/pkg/commands/deploy.go
+++ b/pkg/commands/deploy.go
@@ -17,18 +17,23 @@
 package commands
 
 import (
+	"context"
+	"github.com/fromanirh/deployer/pkg/clientutil"
+	"github.com/fromanirh/deployer/pkg/manifests"
 	"github.com/spf13/cobra"
 )
 
 type deployOptions struct{}
 
 func NewDeployCommand(commonOpts *CommonOptions) *cobra.Command {
-	//	opts := &deployOptions{}
+	opts := &deployOptions{}
 	deploy := &cobra.Command{
 		Use:   "deploy",
 		Short: "deploy the components and configurations needed for topology-aware-scheduling",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// TODO: deploy components here
+			if err := Deploy(opts); err != nil {
+				return err
+			}
 			return nil
 		},
 		Args: cobra.NoArgs,
@@ -37,15 +42,53 @@ func NewDeployCommand(commonOpts *CommonOptions) *cobra.Command {
 }
 
 func NewRemoveCommand(commonOpts *CommonOptions) *cobra.Command {
-	//	opts := &deployOptions{}
+	opts := &deployOptions{}
 	remove := &cobra.Command{
 		Use:   "remove",
 		Short: "remove the components and configurations needed for topology-aware-scheduling",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// TODO: remove components here
+			if err := Remove(opts); err != nil {
+				return err
+			}
 			return nil
 		},
 		Args: cobra.NoArgs,
 	}
 	return remove
+}
+
+func Deploy(opts *deployOptions) error {
+	cs, err := clientutil.New()
+	if err != nil {
+		return err
+	}
+
+	crd, err := manifests.APICRD()
+	if err != nil {
+		return err
+	}
+
+	err = cs.Create(context.TODO(), crd)
+	if err !=nil {
+		return err
+	}
+	return nil
+}
+
+func Remove(opts *deployOptions) error {
+	cs, err := clientutil.New()
+	if err != nil {
+		return err
+	}
+
+	crd, err := manifests.APICRD()
+	if err != nil {
+		return err
+	}
+
+	err = cs.Delete(context.TODO(), crd)
+	if err !=nil {
+		return err
+	}
+	return nil
 }


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/drone/envsubst v1.0.3
 	github.com/spf13/cobra v1.1.1
+	github.com/spf13/pflag v1.0.5
```
fromanirh, line 8: why is this needed?

pkg/commands/consts.go
```diff
@@ -0,0 +1,3 @@
+package commands
+
+const crdDefaultManifestPath = "manifests/api/crd.yaml"
```
fromanirh, line 3: this should be in the `manifests` package

pkg/commands/deploy.go
```diff
@@ -17,35 +17,107 @@
...
 	}
+	opts.AddFlags(remove.Flags())
 	return remove
 }
+
+func (d *deployOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&d.crdPath, "crd-manifest", crdDefaultManifestPath, "crd manifest path")
+}
+
+func (d *deployOptions) Validate() error {
```
fromanirh, line 82: let's not validate: let's just try to read and bail out if it fails.

go.mod
Tal-or: Reading the command flags from the user

fromanirh: I'm not sure it's actually needed, let's check

pkg/commands/deploy.go
```diff
@@ -17,35 +17,107 @@
...
+			}
+
+			if err := opts.Remove(); err != nil {
+				return err
+			}
 			return nil
 		},
 		Args: cobra.NoArgs,
 	}
+	opts.AddFlags(remove.Flags())
```
fromanirh, line 74: what happens if we don't `AddFlags` explicitely?

fromanirh: like, what if we do like https://github.com/fromanirh/deployer/blob/main/pkg/commands/validate.go#L44

Tal-or: I don't know if it's possible let me check

pkg/commands/deploy.go
```diff
@@ -17,35 +17,107 @@
...
+}
+
+func (d *deployOptions) Validate() error {
+	if _, err := ioutil.ReadFile(d.crdPath); errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("crd manifest path %s not exist", d.crdPath)
+	}
+	return nil
+}
+
+func (d *deployOptions) Deploy() error {
```
fromanirh, line 89: this should be a free function if possible:
```
func  Deploy(opts deployOptions) error
```

pkg/commands/deploy.go
```diff
@@ -17,35 +17,107 @@
...
+	}
+
+	err = cs.Create(context.TODO(), crd)
+	if err !=nil {
+		return err
+	}
+	return nil
+}
+
+func (d *deployOptions) Remove() error {
```
fromanirh, line 107: this should be a free function if possible:
```
func  Remove(opts deployOptions) error
```

pkg/commands/render.go
```diff
@@ -48,7 +48,7 @@ func NewRenderCommand(commonOpts *CommonOptions) *cobra.Command {
 func renderManifests(cmd *cobra.Command, commonOpts *CommonOptions, opts *renderOptions, args []string) error {
 	var objs []runtime.Object
 
-	crd, err := manifests.LoadAPICRD()
+	crd, err := manifests.LoadAPICRD(manifests.CrdDefaultManifestPath)
```
fromanirh, line 51: this is a good idea but we need to push it back. The fact we load the manifests is an implementation detail that the consumers of the `manifests` package should not concern with. I'm going to push a rename dropping the `Load` prefix to emphasize this: the `manifests` package will give the client the object, doesn't matter how it gets it.

Later, past MVP, we will add an option to possibly override the builtin manifests.

Tal-or: So basically the client shouldn't be able to provide the path? 
If that's the case i need to remove also the path flag from the command

fromanirh: yes, not now. We need to postpone this feature.

pkg/commands/deploy.go
```diff
@@ -37,15 +42,53 @@ func NewDeployCommand(commonOpts *CommonOptions) *cobra.Command {
...
+				return err
+			}
 			return nil
 		},
 		Args: cobra.NoArgs,
 	}
 	return remove
 }
+
+func Deploy(opts *deployOptions) error {
```
fromanirh, line 60: we will need to reshape this code a bit when the ongoing refactoring (not even a PR posted) of the `deployer` package lands, but for now looks good enough.

There is no logging/explicit feedback of the operations, but we can fix in a later PR.

pkg/manifests/manifests.go
```diff
@@ -36,6 +36,7 @@ const (
 	ComponentAPI                      = "api"
 	ComponentSchedulerPlugin          = "sched"
 	ComponentResourceTopologyExporter = "rte"
+	CrdDefaultManifestPath = "manifests/api/crd.yaml"
```
fromanirh, line 39: the paths should be all changed in a separate PR, but we can fix later.

fromanirh (approved):

fromanirh:
please address the conflicts - if possible, revert the addition of the constant for the CRD path.

Tal-or:
I think you need to explain to me how you vision this whole deployer thingy. 
For example what you mean by "refactoring (not even a PR posted) of the deployer package lands"
It can save us some review cycles.

fromanirh (approved):

fromanirh:
> I think you need to explain to me how you vision this whole deployer thingy.
> For example what you mean by "refactoring (not even a PR posted) of the deployer package lands"
> It might save us some review cycles.

Yeah, the problem is we design as we go, in this stage. But yes, let's sync more offline.

PR #3, opened 2021-07-14
Add code do deploy RTE

fromanirh:
We add support to deploy a topology-updater component, which happens to be RTE for the time being.

### DIFF
--- a/pkg/deployer/rte/rte.go
+++ b/pkg/deployer/rte/rte.go
@@ -38,6 +38,21 @@ type Manifests struct {
 	DaemonSet          *appsv1.DaemonSet
 }
 
+func (mf Manifests) EnforceNamespace() Manifests {
+	ret := Manifests{
+		Namespace: mf.Namespace.DeepCopy(),
+		ServiceAccount: mf.ServiceAccount.DeepCopy(),
+		ClusterRole: mf.ClusterRole.DeepCopy(),
+		ClusterRoleBinding: mf.ClusterRoleBinding.DeepCopy(),
+		DaemonSet: mf.DaemonSet.DeepCopy(),
+	}
+	ret.ServiceAccount.Namespace = ret.Namespace.Name
+	ret.ClusterRole.Namespace = ret.Namespace.Name
+	ret.ClusterRoleBinding.Namespace = ret.Namespace.Name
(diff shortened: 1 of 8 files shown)


pkg/deployer/rte/rte.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+
+	mf, err := GetManifests()
+	if err != nil {
+		return err
+	}
+
+	if err := cli.Create(context.TODO(), mf.Namespace); err != nil {
```
Tal-or: Can we have all those manifests inside a map or something similar and just iterate through that?

fromanirh: Ideally yes, but need to check how. Also, only some manifests may need to be updated, which makes a generic approach a bit more complex.

pkg/deployer/rte/rte.go
```diff
@@ -142,22 +142,20 @@ func Remove(logger *log.Logger, opts Options) error {
 	mf = mf.EnforceNamespace()
 	logger.Printf("manifests loaded")
 
-	// no need to update in the remove path
-	if err := hp.DeleteObject(mf.DaemonSet); err != nil {
+	// since we created everything in the namespace, we can just do
```
Tal-or, line 145: Since the remove logic is heavily relying on the fact that all objects are under the same ns, don't you think it is better to invoke ``` mf.EnforceNamespace()``` inside ```GetManifests()```? this way you keep the user (and by user I mean other contributors) from dropping the call to ```mf.EnforceNamespace()``` and avoid potential issues.

fromanirh: The (admittedly undocumented) intent of `GetManifests` is to provide the pristine versions of the manifests. The trasformations should be explicit as possible, so we should add another helper (and write the docstrings to make the intent expliit) rather then coalesce more responsabilities into `GetManifests`.

I agree about reducing the margin for errors, once we have a more solid design we should indeed add higher level helpers.
In general, catching these errors calls for more tests (unit, e2e) more than else.

So I think we should add automated tests first, let the design solidify a bit meantime, and then revisit the APIs and/or fix the existing functions.

PR #13, opened 2021-07-18
E2E overall deployment

Tal-or:
Run the deployer tools and checks for the following:
- Check that pods are running
- Check that noderesourcetopology object is exist and contain some information


test/log/log.go
```diff
@@ -0,0 +1,23 @@
+package log
```
Tal-or, line 1: This package might be redundant. I added it just because ```NewHelper()``` function is asking for logger

Tal-or:
@fromanirh  Please merge your PR first, since mine depends on yours.
I also see the test found some problems. Will work on that

Tal-or:
/retest
pull the latest ```resource-topology-exporter``` image should fix that

test/e2e/e2e_test.go
```diff
@@ -47,9 +48,45 @@ var _ = ginkgo.BeforeSuite(func() {
 	deployerBaseDir = filepath.Clean(filepath.Join(basedir, "..", ".."))
 	binariesPath = filepath.Clean(filepath.Join(deployerBaseDir, "_out"))
 	fmt.Fprintf(ginkgo.GinkgoWriter, "using binaries at %q\n", binariesPath)
+	deploy()
```
fromanirh, line 51: this means we do deploy after each suite, isnt't it? we should rather do it only in the overall test(s) - for now.

test/e2e/e2e_test.go
```diff
@@ -47,9 +48,45 @@ var _ = ginkgo.BeforeSuite(func() {
...
+})
+
+var _ = ginkgo.AfterSuite(func() {
+	fmt.Fprintf(ginkgo.GinkgoWriter, "clean reasources")
+	remove()
```
fromanirh, line 56: same

test/e2e/e2e_test.go
```diff
@@ -47,9 +48,45 @@ var _ = ginkgo.BeforeSuite(func() {
...
+
+func deploy() {
+	cmdline := []string{
+		filepath.Join(binariesPath, "deployer"),
+		"deploy",
+	}
+	fmt.Fprintf(ginkgo.GinkgoWriter, "running: %v\n", cmdline)
+
+	cmd := exec.Command(cmdline[0], cmdline[1:]...)
+	cmd.Stderr = ginkgo.GinkgoWriter
```
fromanirh, line 72: let's also see the stdout, for debugging purposes:
```
cmd.Stdout = ginkgo.GinkgoWriter
```

test/e2e/e2e_test.go
```diff
@@ -47,9 +48,45 @@ var _ = ginkgo.BeforeSuite(func() {
...
+
+	if _, err := cmd.Output(); err != nil {
```
fromanirh, line 74: now we can do just
```
if err := cmd.Run(); err != nil {
...
```

test/e2e/e2e_test.go
```diff
@@ -47,9 +48,45 @@ var _ = ginkgo.BeforeSuite(func() {
...
+}
+
+func remove() {
+	cmdline := []string{
+		filepath.Join(binariesPath, "deployer"),
+		"remove",
+	}
+	fmt.Fprintf(ginkgo.GinkgoWriter, "running: %v\n", cmdline)
+
+	cmd := exec.Command(cmdline[0], cmdline[1:]...)
```
fromanirh, line 86: same comments as per `deploy`

test/e2e/e2e_test.go
```diff
@@ -47,9 +48,45 @@ var _ = ginkgo.BeforeSuite(func() {
...
+	cmd.Stderr = ginkgo.GinkgoWriter
+
+	if _, err := cmd.Output(); err != nil {
+		fmt.Fprintf(ginkgo.GinkgoWriter, "failed to remove components after test finished: %v", err)
+	}
+}
```
fromanirh, line 92: please add a newline at the end of the file

test/e2e/positive.go
```diff
@@ -17,16 +17,37 @@
 package e2e
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/fromanirh/deployer/pkg/clientutil"
+	corev1 "k8s.io/api/core/v1"
```
fromanirh, line 24: please group the imports properly:

stdlib

kube

3rd party

local project imports

test/e2e/positive.go
```diff
@@ -50,4 +71,98 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+
+	ginkgo.Context("with cluster with all the deployed components", func() {
+
+		ginkgo.It("should have API components", func() {
+			mf, err := api.GetManifests()
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			key := getObjKey(mf.Crd.Name, mf.Crd.Namespace)
+			err = hp.WaitForObjectToBeCreated(key, &v1.CustomResourceDefinition{})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
```
fromanirh, line 83: this must be ensured by `deploy`, so it seems a bit redundant?

test/e2e/positive.go
```diff
@@ -50,4 +71,98 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+
+		ginkgo.It("should have all RTE components", func() {
+			mf, err := rte.GetManifests()
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+			mf = mf.UpdateNamespace()
+
+			for _, obj := range mf.ToObjects() {
+				key := getObjKey(obj.GetName(), obj.GetNamespace())
+				err = hp.WaitForObjectToBeCreated(key, obj)
+				gomega.Expect(err).ToNot(gomega.HaveOccurred(), fmt.Sprintf("object of kind %v was not created", obj.GetObjectKind()))
```
fromanirh, line 94: I'm not sure we need this level of detail at this stage. But if we want to go at this level of detail (and again I don't think we need - now) we should also check the objects all have the right status (e.g. all ready, no errors)

test/e2e/positive.go
```diff
@@ -50,4 +71,98 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+	podNameRgx, err := regexp.Compile(reg)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+	podList := &corev1.PodList{}
+	err = cs.List(context.TODO(), podList)
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	for _, pod := range podList.Items {
+		if match := podNameRgx.FindString(pod.Name); len(match) != 0 {
```
fromanirh, line 163: but we don't actually check the pods are running (and without errors in their status), do we?

test/e2e/positive.go
```diff
@@ -50,4 +71,98 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+			return true
+		}
+	}
+	return false
+}
```
fromanirh, line 168: please add a new line at the end of the file

test/e2e/positive.go
```diff
@@ -50,4 +71,98 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+			gomega.Expect(found).To(gomega.BeTrue())
+		})
+
+		ginkgo.It("should have some information in noderesourcetopology objects", func() {
+			key := getObjKey(nodeResourceTopology, exportedNs)
+			nrt := &v1alpha1.NodeResourceTopology{}
+			err := hp.WaitForObjectToBeDeleted(key, nrt)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			// TODO check if specific resources are exist under the noderesourcetopology object
```
fromanirh, line 135: it's sufficient to check we have > 1 cpus and something (anything) in the topology manager policy.

test/e2e/positive.go
```diff
@@ -50,4 +71,98 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+
+		})
+
+		ginkgo.It("should have all pods running in all namespaces", func() {
+			ginkgo.By("checking that resource-topology-exporter pod is running")
+			mf, err := rte.GetManifests()
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+			mf = mf.UpdateNamespace()
+
+			found := getPodByRegex(fmt.Sprintf("%s-*", mf.DaemonSet.Name))
```
fromanirh, line 117: another option is to take a list of pods in the cluster before `deploy()`, take another after `deploy()` completes succesfully and check the status of the pods found in the latter but not in the former.

test/e2e/positive.go
```diff
@@ -50,4 +71,98 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+			mfs = mfs.UpdateNamespace()
+
+			found = getPodByRegex(fmt.Sprintf("%s-*", mfs.Deployment.Name))
+			gomega.Expect(found).To(gomega.BeTrue())
+		})
+
+		ginkgo.It("should have some information in noderesourcetopology objects", func() {
+			key := getObjKey(nodeResourceTopology, exportedNs)
+			nrt := &v1alpha1.NodeResourceTopology{}
+			err := hp.WaitForObjectToBeDeleted(key, nrt)
```
fromanirh, line 132: not sure "deleted" is correct here

test/log/log.go
fromanirh: I don't have strong opinions, we can keep it.

test/e2e/e2e_test.go
Tal-or: I used ```BeforeAll()``` which is some wrapper to ```BeforeEach()``` that run only once, but I couldn't find any equivalent for that when I wanted to remove everything, IOW ```AfterAll()``` can't work the same. Do you have any other idea?

test/e2e/positive.go
Tal-or: Yes noticed that I've some changes in progress

test/e2e/positive.go
Tal-or: You're right noticed that, have it in progress

test/e2e/e2e_test.go
fromanirh: I'd just do it on `BeforeEach`/`AfterEach` on the new test which checks the correctness of the deploy. If more tests will require this wrapping, we will re-evaluate.

Tal-or: Deploy/Remove for evey new ```It()``` it's kinda heavy don't you think?

fromanirh: Probably. So let's call it explicitely and manually just inside `ginkgo.Context("with cluster with all the deployed components", func() {`

Tal-or: The only drawback here is that we won't be able to remove everything in case of an error in the test. ```defer()``` won't help either

test/e2e/positive.go
Tal-or: Most of it, yes. I'll probably remove it

Tal-or:
@fromanirh 
I have updated most of what we talked about, but when I'm trying to call ```tc, err := nrtupdater.GetTopologyClient("")```
I'm getting the following error:
```  Unexpected error:
        <*errors.errorString | 0xc00034b490>: {
            s: "unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined",
        }
        unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
    occurred
```
working on figure that out

fromanirh:
> @fromanirh
> I have updated most of what we talked about, but when I'm trying to call `tc, err := nrtupdater.GetTopologyClient("")`
> I'm getting the following error:
> 
> ```
>         <*errors.errorString | 0xc00034b490>: {
>             s: "unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined",
>         }
>         unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
>     occurred
> ```
> 
> working on figure that out

This means the code can't load properly the `kubeconfig`, nor can't find automatically one (the latter because it's running inside a cluster, and e2e tests most likely are not). The quickest solution is to export the `KUBECONFIG` environment variable explicitely.
Alternatively, it seems the `controller-runtime/pkg/client` is smart enough to do the right thing automatically:
`https://github.com/fromanirh/deployer/blob/main/pkg/clientutil/client.go#L47`

test/e2e/positive.go
```diff
@@ -50,4 +65,140 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+				}
+
+				return fmt.Errorf("pod %v is not in %v state", pod.Name, corev1.PodRunning)
+			}, 1 * time.Minute, 15 * time.Second).ShouldNot(gomega.HaveOccurred())
+
+			ginkgo.By("checking that noderesourcetopolgy has some information in it")
+			tc, err := clientutil.NewTopologyClient()
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+			gomega.Eventually(func() error {
```
Tal-or, line 121: Here we also need to ```Get()``` the object and play with it, so ```Eventually()``` is not perfect for that job.
What do you think?

fromanirh: It's good enough for now :)

test/e2e/positive.go
```diff
@@ -50,4 +65,140 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer validation", func() {
...
+		})
+
+		ginkgo.It("should perform overall deployment and verify all pods are running", func() {
+			ginkgo.By("checking that resource-topology-exporter pod is running")
+			mf, err := rte.GetManifests()
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+			mf = mf.UpdateNamespace()
+
+			gomega.Eventually(func() error {
+				pod, err := getPodByRegex(fmt.Sprintf("%s-*", mf.DaemonSet.Name))
```
fromanirh, line 87: this work under current CI on which we have exactly one pod because we will have exactly one worker due to resource constraints. But this should be a loop and collect all the pods, be them 1, 2 or 20. Could be fixed in a later PR.

fromanirh:
good enough for now. Thanks for your updates!

PR #35, opened 2021-09-02
NFD as topology updater

Tal-or:
So far we could only use RTE as the topology-updater component.
Now we can replace it with NFD.
NFD is the u/s version of RTE, thus it's is important to have the ability to use it.

The default topology-updater component is RTE since it's more mature and tested for quite some time.

NOTE: The scheduler-plugin version we have doesn't know how to work with NFD, since NFD has a newer API version.


pkg/manifests/yaml/nfd/daemonset.yaml
```diff
@@ -0,0 +1,115 @@
...
+              drop:
+                - ALL
+            readOnlyRootFilesystem: true
+            runAsUser: 0
+          volumeMounts:
+            - mountPath: /host-var/lib/kubelet/config.yaml
+              name: kubelet-podresources-conf
+            - mountPath: /host-var/lib/kubelet/pod-resources/kubelet.sock
+              name: kubelet-podresources-sock
+            - mountPath: /host/sys
```
swatisehgal, line 94: This should be /host-sys

test/e2e/positive.go
```diff
@@ -269,6 +269,56 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer execution", func() {
...
+			err := deploy([]string{"--updater-type", "NFD"})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		})
+
+		ginkgo.AfterEach(func() {
+			err := remove([]string{"--updater-type", "NFD"})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		})
+
+		ginkgo.It("should have some information in the noderesourcetopology object", func() {
```
Tal-or, line 283: This part is a duplicate of https://github.com/k8stopologyawareschedwg/deployer/blob/main/test/e2e/positive.go#L193
but this is only a testing code, so it's not so bad, but I'm open to suggestions.

pkg/commands/deploy.go
```diff
@@ -68,18 +68,20 @@ func NewRemoveCommand(commonOpts *CommonOptions) *cobra.Command {
 			err = sched.Remove(la, sched.Options{
 				Platform:         opts.clusterPlatform,
 				WaitCompletion:   opts.waitCompletion,
-				RTEConfigData:    commonOpts.RTEConfigData,
+				RTEConfigData:    commonOpts.UpdaterConfigData,
```
fromanirh, line 71: this was originally meant as RTE-specific data. Does NFD need some specific configuration? if yes, it makes sense to generalize; if not, I'd not generalize it (yet)

pkg/commands/deploy.go
```diff
@@ -138,8 +140,9 @@ func NewDeploySchedulerPluginCommand(commonOpts *CommonOptions, opts *deployOpti
 			return sched.Deploy(la, sched.Options{
 				Platform:         opts.clusterPlatform,
 				WaitCompletion:   opts.waitCompletion,
-				RTEConfigData:    commonOpts.RTEConfigData,
+				RTEConfigData:    commonOpts.UpdaterConfigData,
```
fromanirh, line 143: Is keeping the name `RTEConfigData` intentional?

pkg/commands/deploy.go
```diff
@@ -158,11 +161,12 @@ func NewDeployTopologyUpdaterCommand(commonOpts *CommonOptions, opts *deployOpti
 			if opts.clusterPlatform == platform.Unknown {
 				return fmt.Errorf("cannot autodetect the platform, and no platform given")
 			}
-			return rte.Deploy(la, rte.Options{
-				Platform:         opts.clusterPlatform,
-				WaitCompletion:   opts.waitCompletion,
-				RTEConfigData:    commonOpts.RTEConfigData,
-				PullIfNotPresent: commonOpts.PullIfNotPresent,
+			return updater.Deploy(la, updater.Options{
```
fromanirh, line 164: another option could be to have different functions (and/or subpackages) taking the same options instead of passing through `UpdaterType`. So we can have, say, `updater/rte` and `updater/nfd`

pkg/commands/deploy.go
```diff
@@ -246,23 +252,26 @@ func deployOnCluster(commonOpts *CommonOptions, opts *deployOptions) error {
...
+		PullIfNotPresent:  commonOpts.PullIfNotPresent,
+		UpdaterType:       commonOpts.UpdaterType,
 	}); err != nil {
 		return err
 	}
 	if err := sched.Deploy(la, sched.Options{
 		Platform:         opts.clusterPlatform,
 		WaitCompletion:   opts.waitCompletion,
-		RTEConfigData:    commonOpts.RTEConfigData,
+		RTEConfigData:    commonOpts.UpdaterConfigData,
```
fromanirh, line 272: is keeping the name `RTEConfigData` intentional?

pkg/commands/render.go
```diff
@@ -60,7 +60,7 @@ func NewRenderAPICommand(commonOpts *CommonOptions, opts *renderOptions) *cobra.
 			if commonOpts.UserPlatform == platform.Unknown {
 				return fmt.Errorf("must explicitely select a cluster platform")
 			}
-			apiManifests, err := api.GetManifests(commonOpts.UserPlatform)
+			apiManifests, err := api.GetManifests(commonOpts.UserPlatform, commonOpts.UpdaterType)
```
fromanirh, line 63: it's suspicious that the API needs to know which updater is going to be used. Smells of convoluted design.

pkg/commands/render.go
```diff
@@ -83,15 +83,15 @@ func NewRenderSchedulerPluginCommand(commonOpts *CommonOptions, opts *renderOpti
 			if err != nil {
 				return err
 			}
-			rteManifests, err := rte.GetManifests(commonOpts.UserPlatform)
+			updaterManifests, err := updater.GetManifestsHandler(commonOpts.UserPlatform, commonOpts.UpdaterType)
```
fromanirh, line 86: why GetManifests*Handler*?

pkg/commands/root.go
```diff
@@ -28,15 +28,16 @@ import (
...
+	Debug             bool
+	UserPlatform      platform.Platform
+	Log               *log.Logger
+	DebugLog          *log.Logger
+	Replicas          int
+	UpdaterConfigData string
+	PullIfNotPresent  bool
+	updaterConfigFile string
+	plat              string
+	UpdaterType       string
```
fromanirh, line 40: private/non-exported fields (`plat`) should be kept last

pkg/commands/root.go
```diff
@@ -66,14 +67,19 @@ func NewRootCommand(extraCmds ...NewCommandFunc) *cobra.Command {
 			// if it is unknown, it's fine
 			commonOpts.UserPlatform, _ = platform.FromString(commonOpts.plat)
 
-			if commonOpts.rteConfigFile != "" {
-				data, err := os.ReadFile(commonOpts.rteConfigFile)
+			if commonOpts.updaterConfigFile != "" {
```
fromanirh, line 70: note that RTE possibly need a separate config file. Does NFD need the same?

pkg/commands/root.go
```diff
@@ -66,14 +67,19 @@ func NewRootCommand(extraCmds ...NewCommandFunc) *cobra.Command {
...
 				if err != nil {
 					return err
 				}
-				commonOpts.RTEConfigData = string(data)
-				commonOpts.DebugLog.Printf("RTE config: read %d bytes", len(commonOpts.RTEConfigData))
+				commonOpts.UpdaterConfigData = string(data)
+				commonOpts.DebugLog.Printf("Updater config: read %d bytes", len(commonOpts.UpdaterConfigData))
 			}
+
+			if commonOpts.UpdaterType != "RTE" && commonOpts.UpdaterType != "NFD" {
```
fromanirh, line 79: these should better be constants in the `updater` package.

pkg/deployer/updater/updater.go
```diff
@@ -0,0 +1,70 @@
...
+
+package updater
+
+import (
+	"fmt"
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
+	"github.com/k8stopologyawareschedwg/deployer/pkg/tlog"
+)
+
+type Type string
```
fromanirh, line 25: why?

pkg/deployer/updater/updater.go
```diff
@@ -0,0 +1,70 @@
...
+
+type Options struct {
+	Platform          platform.Platform
+	WaitCompletion    bool
+	UpdaterConfigData string
+	PullIfNotPresent  bool
+	UpdaterType       string
+}
+
+func Deploy(log tlog.Logger, opts Options) error {
```
fromanirh, line 40: or just create two subpackages and expose the functions directly. Simpler, less code.

pkg/deployer/updater/updater.go
```diff
@@ -0,0 +1,70 @@
...
+	case NFD:
+		return deployNFD(log, nfdOptions{
+			Options:       opts,
+			NFDConfigData: opts.UpdaterConfigData,
+		})
+	}
+	return fmt.Errorf("%s is invalid updater type", opts.UpdaterType)
+}
+
+func Remove(log tlog.Logger, opts Options) error {
```
fromanirh, line 56: same

pkg/images/consts.go
```diff
@@ -20,4 +20,6 @@ const (
 	SchedulerPluginSchedulerDefaultImage  = "k8s.gcr.io/scheduler-plugins/kube-scheduler:v0.19.9"
 	SchedulerPluginControllerDefaultImage = "k8s.gcr.io/scheduler-plugins/controller:v0.19.9"
 	ResourceTopologyExporterDefaultImage  = "quay.io/openshift-kni/resource-topology-exporter:v0.0.19"
+	// TODO replace this image with some image from k8stopologyawareschedwg quay repo
+	NodeFeatureDiscoveryDefaultImage = "quay.io/swsehgal/node-feature-discovery:v0.10.0-devel-36-g5b0761f3"
```
fromanirh, line 24: better not to use developers' repo unless we really don't have options. We need to find out a flow to build images, or to use u/s ones.

pkg/manifests/api/api.go
```diff
@@ -65,13 +66,18 @@ func (mf Manifests) Update() Manifests {
...
 	var err error
 	mf := Manifests{
 		plat: plat,
 	}
 
-	mf.Crd, err = manifests.APICRD()
+	if updaterType == string(updater.NFD) {
+		mf.Crd, err = manifests.APINFDCRD()
+	} else {
+		mf.Crd, err = manifests.APICRD()
```
fromanirh, line 78: this complicates the test matrix too much. A given release of `deployer` shoudl deploy a single, well defined API version.

pkg/manifests/manifests.go
```diff
@@ -147,6 +148,22 @@ func APICRD() (*apiextensionv1.CustomResourceDefinition, error) {
 	return crd, nil
 }
 
+// APINFDCRD will return a newer CRD version that being used by NFD
+// Eventually we should use only the newer version
+// but we need to wait until scheduler plugin and RTE will catch up
+func APINFDCRD() (*apiextensionv1.CustomResourceDefinition, error) {
+	obj, err := loadObject("yaml/api/crd_for_nfd.yaml")
```
fromanirh, line 155: I'm against this. We should bump the APIs, rebuild and only then consume from deployer.

pkg/manifests/updater/updater.go
```diff
@@ -0,0 +1,79 @@
...
+	RTE Type = "RTE"
+	NFD Type = "NFD"
+)
+
+const (
+	namespaceOCP      = "openshift-monitoring"
+	serviceAccountOCP = "node-exporter"
+)
+
+type ManifestsHandler interface {
```
fromanirh, line 45: why is this needed?

pkg/manifests/updates.go
```diff
@@ -164,3 +213,13 @@ func pullPolicy(pullIfNotPresent bool) corev1.PullPolicy {
...
 }
+
+func nfdTopologyUpdaterContainerIndex(containers []corev1.Container) int {
+	for index, cont := range containers {
+		if cont.Name == "nfd-topology-updater" {
+			return index
+		}
+	}
+	// should never happen
+	panic(fmt.Errorf("container named nfd-topology-updater was not found, please check the manifests"))
```
fromanirh, line 224: we should try really, really, really hard to avoid panic in our code

fromanirh (commented):
I think the first patch does too much. The first patch should generalize the updater and add rte as default (and only) option, making room for NFD which should be updated in a later patch.

pkg/manifests/yaml/nfd/daemonset.yaml
```diff
@@ -0,0 +1,115 @@
...
+          name: nfd-topology-updater
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            readOnlyRootFilesystem: true
+            runAsUser: 0
+          volumeMounts:
+            - mountPath: /host-var/lib/kubelet/config.yaml
```
fromanirh, line 90: this won't work on OCP.

pkg/manifests/yaml/nfd/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: node-feature-discovery
```
fromanirh, line 4: we need to check how NFD is deployed on OCP nowadays

test/e2e/positive.go
fromanirh: duplicating test code is actually ok: https://about.sourcegraph.com/go/advanced-testing-in-go/#Golden-files-also-test-flags

test/e2e/positive.go
```diff
@@ -286,14 +336,15 @@ func getNodeResourceTopology(tc *topologyclientset.Clientset, namespace, name st
 	return nrt
 }
 
-func deploy() error {
+func deploy(extraArgs []string) error {
```
fromanirh, line 339: this is too low level. This function (and `remove`) should take an argument to distinguish between NFD and RTE and create the command line accordingly, avoiding to accept random data from the callers.

pkg/commands/deploy.go
Tal-or: The ```RTEConfigData``` contains information such as the Exclude-list, which NFD has as well.

pkg/commands/render.go
Tal-or: This is temporary. Actually, both NFD and RTE should have the same API, but we didn't RTE didn't catch up with NFD latest changes. I also added a comment here about that: https://github.com/k8stopologyawareschedwg/deployer/blob/c8173b1f688e0fa2e04cac34a5c6c836ad8a0a4c/pkg/manifests/manifests.go#L151

Anyway, we can w8 w/ this PR till RTE will sync.

pkg/commands/deploy.go
fromanirh: ack. So yes, let's generalize. Could be worth to rename like `UpdaterSpecificConfig` (also the commandline option should be renamed)

pkg/commands/render.go
fromanirh: yes, we should indeed wait and not mix APIs - especially considering they are unversioned.

pkg/deployer/updater/updater.go
Tal-or: The intention was to keep the API unchanged and hide the implementation details, but ok I'll change it.

fromanirh: in general it's a good thing to care about, but in this case noone is consuming our packages, `deployer` is alpha-level and it is a big change (version-bump worthy) so let's take the simpler approach.
We didn't make any compatibility promise yet, nor we will in the foreseeable future.

pkg/manifests/updater/updater.go
fromanirh: I'm not convinced this bring a real value, I'll review the PR again more carefully.

pkg/images/consts.go
Tal-or: We can have a release flow as RTE has: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/.github/workflows/release.yml

fromanirh: yes, that'sa a good starting point, but it needs few extra steps - and to be setup in our fork :)

pkg/manifests/yaml/api/crd.yaml
```diff
@@ -1,68 +1,144 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    "api-approved.kubernetes.io": "https://github.com/kubernetes/enhancements/pull/1870"
+    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870
```
fromanirh, line 5: so now we bump the API we ship to 0.0.10 AND we add support for NFD as updater, all in this one PR?
If so, this is too much:
- API bump + artifacts bump should be its own PR
- support for NFD as updater should be its own PR (or two, one preparation + one for the actual support)

fromanirh (commented):
looks nice now. Let's clarify the scope of the PR, because the container image references will likely need to be bumped. As it is, this PR doesn't seem completed.
If we split this PR, we can merge some parts of it quite quickly.

pkg/manifests/yaml/api/crd.yaml
Tal-or: OK I'll start split the PRs

Tal-or:
This PR is already stale. it's easier to come up with a new one.

PR #36, opened 2021-09-05
Add extension-apiserver-authentication-reader role to sched

Tal-or:
In the logs of the controller and scheduler pods you can see the following error message:
```
"k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:206: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "extension-apiserver-authentication" is forbidden:
User "system:serviceaccount:tas-scheduler:topology-aware-scheduler" cannot list resource "configmaps" in API group "" in the namespace "kube-system""
```

So we need to add the "extension-apiserver-authentication-reader" in to the scheduler/conttoller's service-account in order to make this error disappear.

We have the same rolebinding here: https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/manifests/noderesourcetopology/cluster-role.yaml#L62 


fromanirh:
besides the warning, what's the consequence of this misconfiguration?

pkg/manifests/sched/sched.go
```diff
@@ -195,6 +205,10 @@ func GetManifests(plat platform.Platform) (Manifests, error) {
 	if err != nil {
 		return mf, err
 	}
+	mf.RBScheduler, err = manifests.RoleBinding(manifests.ComponentSchedulerPlugin, manifests.SubComponentSchedulerPluginScheduler)
```
fromanirh, line 208: do we need this RoleBinding in both k8s and ocp flows or only in the former?

pkg/manifests/yaml/sched/controller/rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: topology-aware-controller-as-kube-controller
+  namespace: kube-system
+subjects:
+  - kind: ServiceAccount
+    name: topology-aware-controller
+    namespace: tas-scheduler
```
fromanirh, line 9: do we need to adjust the namespace(s) here? - `pkg/manifests/updates.go`

pkg/manifests/manifests.go
```diff
@@ -134,6 +134,26 @@ func ClusterRoleBinding(component, subComponent string) (*rbacv1.ClusterRoleBind
 	return crb, nil
 }
 
+func RoleBinding(component, subComponent string) (*rbacv1.RoleBinding, error) {
+	if err := validateComponent(component); err != nil {
```
fromanirh, line 138: shouldn't this be allowed only for scheduler plugins? (component == "sched")

Tal-or:
> besides the warning, what's the consequence of this misconfiguration?

Initially, I thought it is the reason for pods getting stuck in a pending state, but it was not.
So it's just the warnning message

pkg/manifests/sched/sched.go
Tal-or: I saw the warning on both components.
In general, the controller is an OCP flow?

pkg/manifests/yaml/sched/controller/rolebinding.yaml
Tal-or: It can be part of ```kube-system``` ns, but it's not mandatory I think.
I don't see ```pkg/manifests/updates.go``` file under this rep, what am I missing here?

pkg/manifests/manifests.go
Tal-or: Yes. but that's what i'm doing 
```mf.RBController, err = manifests.RoleBinding(manifests.ComponentSchedulerPlugin, manifests.SubComponentSchedulerPluginController)```

fromanirh:
let's check if https://github.com/kubernetes-sigs/scheduler-plugins/issues/271 and https://github.com/kubernetes-sigs/scheduler-plugins/pull/272 are relevant and/or how we can consume the proposed fixes.

Tal-or:
> let's check if [kubernetes-sigs/scheduler-plugins#271](https://github.com/kubernetes-sigs/scheduler-plugins/issues/271) and [kubernetes-sigs/scheduler-plugins#272](https://github.com/kubernetes-sigs/scheduler-plugins/pull/272) are relevant and/or how we can consume the proposed fixes.

Sure. I'll get to that later today, or tomorrow

Tal-or:
> let's check if [kubernetes-sigs/scheduler-plugins#271](https://github.com/kubernetes-sigs/scheduler-plugins/issues/271) and [kubernetes-sigs/scheduler-plugins#272](https://github.com/kubernetes-sigs/scheduler-plugins/pull/272) are relevant and/or how we can consume the proposed fixes.

Basically, this is identical to what was suggested in this PR.
But we won't be able to enjoy this fix since our deployment method is different unless we'll consume the same manifest via vendoring.
Is that something that we want to do?

Tal-or:
> > let's check if [kubernetes-sigs/scheduler-plugins#271](https://github.com/kubernetes-sigs/scheduler-plugins/issues/271) and [kubernetes-sigs/scheduler-plugins#272](https://github.com/kubernetes-sigs/scheduler-plugins/pull/272) are relevant and/or how we can consume the proposed fixes.
> 
> Basically, this is identical to what was suggested in this PR. But we won't be able to enjoy this fix since we're using a different set of manifests. Unless we'll consume the same manifest via vendoring. Is that something that we want to do?

@fromanirh ^

fromanirh:
no, we don't want to vendor manifests (usually only code is vendored). So let's go forward with this PR

Tal-or:
After rebase I see no changes. It looks like the PR already go merged https://github.com/k8stopologyawareschedwg/deployer/commit/a21d60126a949471c82b89281ae4c3ace02f4178 by mistake

PR #43, opened 2021-10-08
rte: use role, not clusterrole

ffromani:
There is no strong reason to use a clusterrole{,binding} for RTE,
while using role{,binding} would make the resources used by RTE
almost entirely confined in a NS, thus easier to track and to remove.

No intended change in behaviour.

### DIFF
--- a/pkg/manifests/manifests_test.go
+++ b/pkg/manifests/manifests_test.go
@@ -101,6 +101,102 @@ func TestGetServiceAccount(t *testing.T) {
 	}
 }
 
+func TestGetRole(t *testing.T) {
+	type testCase struct {
+		component    string
+		subComponent string
+		expectError  bool
+	}
+
+	testCases := []testCase{
+		{
+			component:   "unknown-wrong",
+			expectError: true,
+		},
+		{
+			component:   ComponentAPI,
+			expectError: true,
+		},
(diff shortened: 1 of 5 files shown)


pkg/manifests/manifests_test.go
```diff
@@ -101,6 +101,102 @@ func TestGetServiceAccount(t *testing.T) {
...
+			} else {
+				if err != nil || obj == nil {
+					t.Fatalf("nil obj or non-nil err=%v", err)
+				}
+			}
+		})
+	}
+}
+
 func TestGetClusterRole(t *testing.T) {
```
swatisehgal, line 200: Do we still need this?

swatisehgal (commented):
The PR looks good to me but I was wondering why we are still keeping the tests pertaining to the `ClusterRole` inspite of the move to `Role` (and `RoleBinding`). Maybe we should remove it here or remove in a subsequent PR?

pkg/manifests/manifests_test.go
fromanirh: no, better to get rid of them now.

fromanirh:
we actually still need Clusterole{,Binding} fro the scheduling side, so merging this change as-is.

PR #49, opened 2021-10-16
manifests: object state reporting

fromanirh:
We aim to provide the foundations needed to support a reconcile loop using the `pkg/manifests` code.

ObjectState is a umbrella term to implement manifest updates. We add facilities to extract the manifest state and to provide
foundations ofr a reconcile loop. 

The code will provide
- status of the current system: existing manifests (+ error in fetching   them, if any)
- status of the desired state of the system
- helper functions to compare manifests (existing vs desired)
- helper functions to merge manifests, which will be used in the
  reconciliation loops.

### DIFF
--- a/pkg/clientutil/client.go
+++ b/pkg/clientutil/client.go
@@ -42,8 +42,8 @@ func New() (client.Client, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


fromanirh:
no longer needed

PR #54, opened 2021-11-08
Provide and install OpenShift related manifests

cynepco3hahue:
- install MachineConfig to provide a new SELinux policy
- install SecurityContextConstraint to provide RTE DaemonSet additional permissions


pkg/manifests/yaml/rte/securitycontextconstrain.yaml
```diff
@@ -0,0 +1,16 @@
+# TODO: check if we can reduce scc capabilities
```
fromanirh, line 1: please rename to `securitycontextconstraint.yaml` (note the last "t" before `.yaml`)

pkg/commands/render.go
```diff
@@ -133,7 +133,7 @@ func makeRTEObjects(commonOpts *CommonOptions) ([]client.Object, string, error)
 	if err != nil {
 		return nil, namespace, err
 	}
-	mf = mf.Update(rtemanifests.UpdateOptions{
+	mf = mf.Update(&rtemanifests.UpdateOptions{
```
fromanirh, line 136: why the pointer?

pkg/manifests/manifests.go
```diff
@@ -254,6 +257,38 @@ func DaemonSet(component string) (*appsv1.DaemonSet, error) {
 	return ds, nil
 }
 
+func MachineConfig(component string) (*machineconfigv1.MachineConfig, error) {
+	if err := validateComponent(component); err != nil {
```
fromanirh, line 261: this is too broad: both the MachineConfig and the SecurityContext make sense only for `component==rte`.
But can be fixed on a later PR

pkg/manifests/rte/rte.go
```diff
@@ -31,62 +32,93 @@ import (
...
+	// Config Map options
+	ConfigData string
+
+	// General options
+	Namespace string
+	Name      string
 }
 
-func (mf Manifests) Update(options UpdateOptions) Manifests {
+func (mf *Manifests) Update(options *UpdateOptions) *Manifests {
```
fromanirh, line 85: we intentionally returned a full copy (not a pointer) to emphasize the function was not modifying the original data

pkg/manifests/rte/rte.go
```diff
@@ -165,25 +197,33 @@ func (mf Manifests) ToDeletableObjects(hp *deployer.Helper, log tlog.Logger) []d
 	return objs
 }
 
-func New(plat platform.Platform) Manifests {
-	mf := Manifests{
+func New(plat platform.Platform) *Manifests {
```
fromanirh, line 200: I'm still not sold about the pointer. The intent of the previous function was indeed to suggest the the functions were not changing the object - this is why the DeepCopy() and friends.

fromanirh (commented):
initial review, need to spend more time about the SCC changes

pkg/manifests/rte/rte.go
cynepco3hahue: Like I can see that all k8s DeepCopy methods generated in the form of
```
func (in *Policy) DeepCopy() *Policy {
	if in == nil {
		return nil
	}
	out := new(Policy)
	in.DeepCopyInto(out)
	return out
}
```

fromanirh: sure, that part is fine. My point is more about the methods of the `Manifests` object.
When you have a pointer receiver, you get the message "this method may mutate the receiver object". This is not the case by design about the Manifests object. if we move to pointer receiver we lose this information, every method is potentially mutating even if actually none is (or should). And I'm not sure at all the gains we get are worthy.

pkg/manifests/rte/rte.go
cynepco3hahue: revert it back

pkg/manifests/manifests.go
cynepco3hahue: do you want to move it under the `rte` package?

pkg/manifests/yaml/rte/securitycontextconstrain.yaml
cynepco3hahue: done

pkg/manifests/manifests.go
fromanirh: no need, just drop the call to `validateComponent` and use a plain
```
if component == rte {
...
```

pkg/manifests/rte/rte.go
cynepco3hahue: Under additional thinking, let leave this one with a pointer to avoid the creation of the new object during the assignment.

pkg/commands/render.go
cynepco3hahue: same as with new, we will create two objects if we will not pass the pointer:
1. rtemanifests.UpdateOptions{...}
2. copy of rtemanifests.UpdateOptions{...} under the method

pkg/deployer/rte/rte.go
```diff
@@ -108,18 +106,15 @@ func Remove(log tlog.Logger, opts Options) error {
 		return err
 	}
 	namespace := ""
-	if opts.Platform == platform.Kubernetes {
+	if opts.Platform == platform.Kubernetes || opts.Platform == platform.OpenShift {
```
Tal-or, line 109: dito

pkg/deployer/rte/rte.go
```diff
@@ -37,16 +37,14 @@ type Options struct {
 }
 
 func SetupNamespace(plat platform.Platform) (*corev1.Namespace, string, error) {
-	if plat == platform.Kubernetes {
+	if plat == platform.Kubernetes || plat == platform.OpenShift {
```
Tal-or, line 40: I guess this line can be omitted, we don't have another platform, even if it's unknown, it can be only one of the two options.

pkg/manifests/updates.go
```diff
@@ -89,14 +97,18 @@ func UpdateSchedulerConfigNamespaces(logger tlog.Logger, cm *corev1.ConfigMap, N
...
 	// TODO: better match by name than assume container#0 is RTE proper (not minion)
 	ds.Spec.Template.Spec.Containers[0].Image = images.ResourceTopologyExporterImage
 	ds.Spec.Template.Spec.Containers[0].ImagePullPolicy = pullPolicy(pullIfNotPresent)
 	if len(ds.Spec.Template.Spec.Containers) > 1 {
 		// TODO: more polite/proper iteration
 		ds.Spec.Template.Spec.Containers[1].ImagePullPolicy = pullPolicy(pullIfNotPresent)
 	}
+
+	// TODO: better match by name than assume container#0 is RTE proper (not minion)
+	ds.Spec.Template.Spec.Containers[0].Image = images.ResourceTopologyExporterImage
```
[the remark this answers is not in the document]
cynepco3hahue: Rebase issue

pkg/deployer/rte/rte.go
cynepco3hahue: yee, I think we can.

go.mod
```diff
@@ -7,18 +7,19 @@ require (
 	github.com/hashicorp/go-version v1.2.0
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.0.10
 	github.com/onsi/ginkgo v1.16.4
-	github.com/onsi/gomega v1.13.0
-	github.com/openshift/api v0.0.0-20210713130143-be21c6cb1bea // indirect
-	github.com/openshift/client-go v0.0.0-20200320143156-e7fa42a1261e
-	github.com/spf13/cobra v1.1.3
+	github.com/onsi/gomega v1.14.0
+	github.com/openshift/api v0.0.0-20210924154557-a4f696157341
+	github.com/openshift/client-go v0.0.0-20210916133943-9acee1a0fb83
```
fromanirh, line 12: this is fine. Do we need to take steps to make sue this is compatible with the k8s client-go we are using?

pkg/manifests/yaml/rte/machineconfig.yaml
```diff
@@ -0,0 +1,33 @@
+# To update the SELinux policy you will need to decode the source content via base64, update it and encode back
```
fromanirh, line 1: so it seems better to have it empty - or with a well-known and documented placeholder, and fill it programmatically.

pkg/commands/render.go
fromanirh: let's postpone this change in another PR. Let's keep copy by value for the time being.

pkg/deployer/rte/rte.go
fromanirh: yep. Let's drop this `if` entirely.

pkg/manifests/manifests.go
```diff
@@ -254,6 +257,38 @@ func DaemonSet(component string) (*appsv1.DaemonSet, error) {
 	return ds, nil
 }
 
+func MachineConfig(component string) (*machineconfigv1.MachineConfig, error) {
+	if err := validateRTEComponent(component); err != nil {
```
fromanirh, line 261: let's inline this check:
```
if component != ComponentResourceTopologyExporter {
	return fmt.Errorf("component %q is not an %q component", component, ComponentResourceTopologyExporter)
}
```

pkg/manifests/manifests.go
```diff
@@ -254,6 +257,38 @@ func DaemonSet(component string) (*appsv1.DaemonSet, error) {
...
+
+	mc, ok := obj.(*machineconfigv1.MachineConfig)
+	if !ok {
+		return nil, fmt.Errorf("unexpected type, got %t", obj)
+	}
+	return mc, nil
+}
+
+func SecurityContextConstraint(component string) (*securityv1.SecurityContextConstraints, error) {
+	if err := validateRTEComponent(component); err != nil {
```
fromanirh, line 277: ditto

pkg/manifests/rte/rte.go
fromanirh: I prefer to rever this change, please. If we do this change, and that will require discussion, we will need to update the other subpackages (`pkg/manifests/{sched,api}`) to have a consistent codebase.

fromanirh (commented):
need to review the fine details of the selinux part, but overall looks good

pkg/deployer/rte/rte.go
cynepco3hahue: done

pkg/deployer/rte/rte.go
cynepco3hahue: done

pkg/manifests/updates.go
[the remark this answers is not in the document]
cynepco3hahue: done

go.mod
cynepco3hahue: it updated automatically by `go mod tidy` so I think it good to go, otherwise you will have errors under func or unit tests

pkg/manifests/yaml/rte/machineconfig.yaml
cynepco3hahue: So we have two options:
1. save it under the const or variable under the code, IMHO it good enough 
2. save it under the file, we will need to copy it under the container image, and inject it dynamically during the start

pkg/manifests/manifests.go
cynepco3hahue: done

pkg/manifests/manifests.go
cynepco3hahue: done

pkg/manifests/yaml/rte/machineconfig.yaml
fromanirh: Let's go with the first option for now, seems good enough indeed

go.mod
fromanirh: works for me!

pkg/manifests/updates.go
```diff
@@ -154,6 +184,105 @@ func UpdateResourceTopologyExporterCommand(args []string, vars map[string]string
 	return res
 }
 
+func UpdateMachineConfig(mc *machineconfigv1.MachineConfig, name string, mcpSelector *metav1.LabelSelector) error {
+	if name != "" {
+		mc.Name = fmt.Sprintf("51-%s", name)
+	}
+
+	ignitionConfig, err := getIgnitionConfig()
```
fromanirh, line 192: ok, this is why the error surfaces up to `Update()`. We should load all the assets we need explicitely. Moving this code into `GetManifests` is still a stretch (selinux policies and ignition config don't really quality as `manifests`) but since they are required for the MachineConfig **manifest** they still kinda fit into the picture.

This cleanup can be done later, though.

fromanirh (commented):
Few not-so-minor things. Good enough to move forward, from my POV we can iterate later.

fromanirh:
/lgtm
see the inline comments

pkg/manifests/updates.go
```diff
@@ -154,6 +184,105 @@ func UpdateResourceTopologyExporterCommand(args []string, vars map[string]string
...
+
+	rawIgnition, err := json.Marshal(ignitionConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	return rawIgnition, nil
+}
+
+func getSELinuxInstallSystemdServiceContent() (*string, error) {
```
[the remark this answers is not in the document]
cynepco3hahue: because ignition get the string pointer, so I prefer just to return *string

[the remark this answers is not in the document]
fromanirh: ok, no bother. Just document this in the function docstring and we're good.

fromanirh (commented):
spotted a possible bug, otherwise we are good to go.

pkg/deployer/rte/rte.go
```diff
@@ -107,13 +99,7 @@ func Remove(log tlog.Logger, opts Options) error {
...
 		return err
 	}
-	namespace := ""
-	if opts.Platform == platform.Kubernetes {
-		namespace = ns.Name
-	}
-	if opts.Platform == platform.OpenShift {
-		namespace = rtemanifests.NamespaceOpenShift
-	}
+	namespace := ns.Namespace
```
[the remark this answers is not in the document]
cynepco3hahue: nice catch!

pkg/manifests/manifests.go
```diff
@@ -19,22 +19,28 @@ package manifests
 import (
 	"bytes"
 	"embed"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"html/template"
```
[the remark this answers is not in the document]
cynepco3hahue: it should, done

pkg/manifests/manifests.go
```diff
@@ -254,6 +262,120 @@ func DaemonSet(component string) (*appsv1.DaemonSet, error) {
...
+
+	rawIgnition, err := json.Marshal(ignitionConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	return rawIgnition, nil
+}
+
+func getSELinuxInstallSystemdServiceContent() (*string, error) {
```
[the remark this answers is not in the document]
cynepco3hahue: done

fromanirh (approved):
thanks, looks really nice!
In the future we will need to improve the test coverage, and/or add CI coverage for OCP like we do for k8s. Unfortunately, I'm not aware of solutions simple as kind to quickly setup OCP clisters, so having CI will be more challenging here.

PR #56, opened 2021-11-11
Manifests.Update() will be called on fresh DaemonSet copy

Tal-or:
The deployer's buisness logic was intended to get executed once per the deployer process's lifetime.
We reused the same logic on the numaresources-operator, but on the operator we're performing multiple updates on the same manifests instance - IOW the Update() function processed a dirty copy of DaemonSet object if more than a single Update() call being executed.

This lead to several bugs:
1. https://github.com/k8stopologyawareschedwg/deployer/pull/55
2. The export-namespace https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/manifests/yaml/rte/daemonset.yaml#L20 command argument, not updated properly (the initial reason for PR)

In order to solve this issue we change the Update() logic to always work on a fresh copy of DaemonSet object, hence we won't carry former changes during the Update() call.

### DIFF
--- a/pkg/manifests/rte/rte.go
+++ b/pkg/manifests/rte/rte.go
@@ -45,6 +45,8 @@ type Manifests struct {
 	// internal fields
 	plat           platform.Platform
 	serviceAccount string
+	// DaemonSet which not processed by Manifests.Update()
+	initialDaemonSet *appsv1.DaemonSet
 }
 
 func (mf Manifests) Clone() Manifests {
@@ -52,9 +54,10 @@ func (mf Manifests) Clone() Manifests {
 		plat:           mf.plat,
 		serviceAccount: mf.serviceAccount,
 		// objects
-		Role:        mf.Role.DeepCopy(),
-		RoleBinding: mf.RoleBinding.DeepCopy(),
-		DaemonSet:   mf.DaemonSet.DeepCopy(),
+		Role:             mf.Role.DeepCopy(),
+		RoleBinding:      mf.RoleBinding.DeepCopy(),
+		DaemonSet:        mf.DaemonSet.DeepCopy(),
+		initialDaemonSet: mf.initialDaemonSet.DeepCopy(),
 	}
 	if mf.plat == platform.Kubernetes {
 		ret.ServiceAccount = mf.ServiceAccount.DeepCopy()
@@ -70,6 +73,8 @@ type UpdateOptions struct {
 
 func (mf Manifests) Update(options UpdateOptions) Manifests {
 	ret := mf.Clone()
+	// work on a fresh (not dirty) clone of DaemonSet
+	ret.DaemonSet = ret.initialDaemonSet.DeepCopy()
 	if ret.plat == platform.Kubernetes {
 		if options.Namespace != "" {
 			ret.ServiceAccount.Namespace = options.Namespace
@@ -197,5 +202,7 @@ func GetManifests(plat platform.Platform) (Manifests, error) {
 	if err != nil {
 		return mf, err
 	}
+
+	mf.initialDaemonSet = mf.DaemonSet.DeepCopy()
 	return mf, nil
 }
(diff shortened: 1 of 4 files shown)


pkg/manifests/rte/rte.go
```diff
@@ -45,16 +45,19 @@ type Manifests struct {
...
 		plat:           mf.plat,
 		serviceAccount: mf.serviceAccount,
 		// objects
-		Role:        mf.Role.DeepCopy(),
-		RoleBinding: mf.RoleBinding.DeepCopy(),
-		DaemonSet:   mf.DaemonSet.DeepCopy(),
+		Role:             mf.Role.DeepCopy(),
+		RoleBinding:      mf.RoleBinding.DeepCopy(),
+		DaemonSet:        mf.DaemonSet.DeepCopy(),
+		initialDaemonSet: mf.initialDaemonSet.DeepCopy(),
```
fromanirh, line 60: this is workarounding a deeper bug. `mf.DaemonSet` is indeed supposed to be the initial pristine copy. We should never mutate that copy. This is why `Update` begins with `ret := mf.Clone()`.
Where in the flow we mutate the original reference of the DaemonSet?

fromanirh (commented):
Good catch. This is indeed a real (bad) bug, indeed hidden by the fact the deployer runs Update()  once. However, the code was designed to be able to support multiple calls to `Update()`.
This means the original code is indeed buggy (great that you are adding the tests), but this also mean there is a deeper bug somewhere in the update flow, because we should never be mutating the original objects.

pkg/manifests/rte/rte.go
Tal-or: hmm interesting let me look deeper.

Tal-or: Damn! :) https://github.com/openshift-kni/numaresources-operator/blob/main/controllers/numaresourcesoperator_controller.go#L117

fromanirh: nevertheless, the bug is real. The shell expansion in the DS command happens only once, and this is not the original intent.

Tal-or: I couldn't come up with a better patch that addresses the shell expansion specifically.
If you have an idea I'm open to suggestions.

fromanirh: After more closer inspection, `Clone()` and `Update()` work as intended. `DeepCopy()` is doing the right thing.
The bug is real, but is in how the manifests are consumed and not in how we generate and update them.

The only real bug here is in the naming: most of the functions the original author ( :) ) called `Update` aren't updating, rather transforming the manifests. "update" is typically in-place, and here we go to some great extent to avoid in-place transformations.

In the future we should rename the code with more explicative names.

Tal-or: I agree. This bug is more about wrong using than wrong implementation.

Anyway let me know what you prefer.
Just merging the unittest from this PR and remove the rest, is also an option.

fromanirh: The test addition is a good idea, but let's file a new PR later in the future just adding tests.

pkg/manifests/rte/rte.go
```diff
@@ -70,6 +73,8 @@ type UpdateOptions struct {
 
 func (mf Manifests) Update(options UpdateOptions) Manifests {
 	ret := mf.Clone()
```
cynepco3hahue, line 75: IMHO it's the bug in the `mf.Clone` method, it should just deep copy all entries under the initial struct

pkg/manifests/rte/rte.go
fromanirh: Let's add tests to verify that the behaviour of `Clone()` and `Update()` - which should never alter the original reference, but always return a mutated copy - is indeed correct. Of course let's do in a new PR.

Tal-or: Sure. I Will issue PR soon.

Tal-or:
As @fromanirh suggested: Let's add tests to verify that the behaviour of Clone() and Update() - which should never alter the original reference, but always return a mutated copy - is indeed correct. Of course let's do in a new PR.

PR #57, opened 2021-11-14
Add unit-tests for Clone and Update

Tal-or:
These unit-tests are making sure that:
1. Clone creates a separate Manifests copy.
2. Update returns a separate Manifests copy and not changing the original one.


fromanirh (approved):
/lgtm
this is a nice start

PR #58, opened 2021-11-15
Provide RTE notifier OCI hook

cynepco3hahue:
To reduce permissions required to monitor files under the `/var/lib/kubelet` the code will allow monitoring the file under the `/run/rte/notify`.
Currently, it will work only for the OpenShift platform. The code will deploy an OCI hook that will touch the specified file under the host each time the guaranteed pod started, which will initiate an update of the NRT object without the need to wait for a sample time.


pkg/assets/rte/rtenotifier.template
```diff
@@ -0,0 +1,18 @@
...
+
+bundle=$(jq '.bundle' /dev/stdin 2>&1)
+pod_cgroup=$(cat ${bundle}/config.json | jq .linux.cgroupsPath)
+
+# Do not touch file for non guaranteed pods
+if [[ ${pod_cgroup} ~= ".*burstable.*" ]] || [[ ${pod_cgroup} ~= ".*besteffort.*" ]]; then
+    exit 0
+fi
+
+notify_file_dir=$(dirname {{.notifierFilePath}})
```
fromanirh, line 13: this is nice, but I'd just try to touch and if the path doesn't exist, good riddance:
```
touch {{ .notifierFilepath }} || :
```

pkg/assets/rte/rtenotifier.template
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -e
+
+bundle=$(jq '.bundle' /dev/stdin 2>&1)
+pod_cgroup=$(cat ${bundle}/config.json | jq .linux.cgroupsPath)
```
fromanirh, line 6: let's save a process spawn: 
```
pod_cgroup=$(jq .linux.cgroupsPath < ${bundle}/config.json)
```

pkg/manifests/manifests.go
```diff
@@ -291,22 +296,103 @@ func DaemonSet(component string, plat platform.Platform, namespace string) (*app
...
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/var/lib/kubelet/pod-resources/kubelet.sock",
+					Type: &hostPathSocket,
+				},
+			},
+		},
+	}
+
+	containerPodResourcesSocket := fmt.Sprintf("/%s/kubelet.sock", rtePodresourcesSocketVolumeName)
```
fromanirh, line 323: let's use `filepath.Join` please

fromanirh: here and in all the similar places (if any)

pkg/manifests/manifests.go
```diff
@@ -376,7 +468,7 @@ func getIgnitionConfig() ([]byte, error) {
 	// load RTE notifier script
 	notifierScript, err := getTemplateContent(rteassets.RTENotifierScript, map[string]string{
 		// TODO: we should use the variable passed from the deployer or operator
-		templateNotifierFilePath: "/run/rte/notify",
+		templateNotifierFilePath: fmt.Sprintf("/%s/%s", rteNotifierVolumeName, rteNotifierFileName),
```
fromanirh, line 471: like here, `filepath.Join` please

fromanirh (commented):
the overall approach lgtm
will have a deeper review later

pkg/manifests/manifests.go
cynepco3hahue: done

pkg/manifests/manifests.go
cynepco3hahue: done

fromanirh:
please add some context in the PR description

pkg/manifests/manifests.go
```diff
@@ -351,11 +353,47 @@ func MachineConfig(component string) (*machineconfigv1.MachineConfig, error) {
...
+	files = addFileToIgnitionConfig(
+		files,
+		notifierHookConfigContent,
+		644,
+		filepath.Join(defaultOCIHooksDir, "rte-notifier.json"),
+	)
+
+	// load RTE notifier script
+	notifierScript, err := getTemplateContent(rteassets.RTENotifierScript, map[string]string{
+		// TODO: we should use the variable passed from the deployer or operator
```
fromanirh: yes, this should be a parameter from the deployer that is passed through up to this code

pkg/manifests/manifests.go
```diff
@@ -60,7 +60,7 @@ const (
 	defaultIgnitionVersion       = "3.2.0"
 	defaultIgnitionContentSource = "data:text/plain;charset=utf-8;base64"
 	defaultOCIHooksDir           = "/etc/containers/oci/hooks.d"
-	defaultScriptsDir            = "/usr/bin"
+	defaultScriptsDir            = "/usr/local/bin"
```
fromanirh, line 63: doesn't really matter, so I'm OK with this

pkg/manifests/manifests.go
```diff
@@ -70,7 +70,13 @@ const (
...
 
 const (
-	containerNameRTE = "resource-topology-exporter"
+	containerNameRTE                = "resource-topology-exporter"
+	rteNotifierVolumeName           = "host-run-rte"
+	rteSysVolumeName                = "host-sys"
+	rtePodresourcesSocketVolumeName = "host-podresources-socket"
+	rteKubeletDirVolumeName         = "host-var-lib-kubelet"
+	rteNotifierFileName             = "notify"
+	hostNotifierDir                 = "/run/rte"
```
fromanirh, line 79: this should be overridable with a parameter as well (in the future)

pkg/manifests/manifests_test.go
```diff
@@ -426,16 +432,146 @@ func TestGetDaemonSet(t *testing.T) {
...
+				if !ok {
+					t.Fatalf("the volume mount %q does not exist under expected volumes mounts %v", m.Name, tc.expectedVolumeMounts)
+				}
+
+				if m.MountPath != path {
+					t.Fatalf("the volume mount %q path %q does not have expected value %q", m.Name, m.MountPath, path)
+				}
+			}
+
+			containerCommand := strings.Join(rteContainer.Command, " ")
```
fromanirh, line 541: this is a bit fragile, but I don't have better suggestions so let's keep like this for the time being

fromanirh:
overall looks good, we can improve few areas but I'd rather do incrementally. Basically once we have a good PR description for historical record, we can merge

pkg/manifests/manifests.go
cynepco3hahue: looks like it matters, it tried to write it before to `/usr/bin`, and the machine config said that it is the read-only directory.

pkg/manifests/manifests.go
cynepco3hahue: For now, I prefer to hardcore it because otherwise, we can not guarantee that the code will work on top of the OpenShift, because of the SELinux. For the hardcoded path we know what SELinux permissions grant to the container.

pkg/manifests/manifests.go
cynepco3hahue: ye, see my comment above, I just unsure regarding to the SELinux.

pkg/manifests/manifests.go
fromanirh: ok, good point. I was thinking only from the FileSystem Hierarchy Standard perspective.

pkg/manifests/manifests.go
fromanirh: Works for me, but let's document this in a code comment.

pkg/manifests/manifests.go
fromanirh: same, this deserves a comment above `defaultScriptsDir`

fromanirh:
actually let's merge, I want to propose an improvement to start generalizing the work done here

PR #59, opened 2021-11-15
Add missing logic for updating metrics port

Tal-or:
The deployer is not updating the DaemonSet with the necessary configuration for support metrics exports
as we have in the RTE repo: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/master/manifests/resource-topology-exporter-ds.yaml#L42

### DIFF
--- a/pkg/manifests/updates.go
+++ b/pkg/manifests/updates.go
@@ -2,6 +2,7 @@ package manifests
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/drone/envsubst"
 	securityv1 "github.com/openshift/api/security/v1"
@@ -26,6 +27,8 @@ const (
 	seLinuxRTEContextLevel = "s0"
 
 	templateSELinuxPolicyDst = "selinuxPolicyDst"
+
+	metricsPort = 2112
 )
 
 func UpdateRoleBinding(rb *rbacv1.RoleBinding, serviceAccount, namespace string) *rbacv1.RoleBinding {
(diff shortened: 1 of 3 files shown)


fromanirh (commented):
good catch. This is a bug indeed. I think that rather than just adding special-purpose code, we should re-evaluate how we handle the shell expansion.

Tal-or:
> good catch. This is a bug indeed. I think that rather than just adding special-purpose code, we should re-evaluate how we handle the shell expansion.

Let me think about it.
Will post something tomorrow

fromanirh:
> > good catch. This is a bug indeed. I think that rather than just adding special-purpose code, we should re-evaluate how we handle the shell expansion.
> 
> Let me think about it. Will post something tomorrow

Let's address the point tomorrow indeed. I'm inclined to merge your PR and schedule a cleanup sweep later on. All the `Update` function  and `updates.go` in general can use some cleanup.

Tal-or:
> > > good catch. This is a bug indeed. I think that rather than just adding special-purpose code, we should re-evaluate how we handle the shell expansion.
> > 
> > 
> > Let me think about it. Will post something tomorrow
> 
> Let's address the point tomorrow indeed. I'm inclined to merge your PR and schedule a cleanup sweep later on. All the `Update` function and `updates.go` in general can use some cleanup.

If you're too busy, but have some idea in my mind, you can share it and i'll implement what's needed

PR #60, opened 2021-11-16
manifests: overhaul of updates

fromanirh:
fix some accumulated cruft in the update code path for better readability and to simplify the flow.


fromanirh:
depends on: https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/74

Tal-or (commented):
This is an API change and will require a follow-up PR in packages that consume the deployer - numaresources-operator

fromanirh:
> This is an API change and will require a follow-up PR in packages that consume the deployer - numaresources-operator

yep, absolutely. When the projects which consume this package bump their `go.mod`, that change is in order.

pkg/manifests/yaml/rte/daemonset.yaml
```diff
@@ -14,11 +14,10 @@ spec:
       serviceAccountName: rte
       containers:
       - name: resource-topology-exporter-container
-        image: ${RTE_CONTAINER_IMAGE}
+        image: quay.io/k8stopologyawareschedwg/resource-topology-exporter:latest
         command:
         - /bin/resource-topology-exporter
-        - --export-namespace=${EXPORT_NAMESPACE}
```
Tal-or, line 20: What about ```--export-namespace```? I guess that now it will be the ```default``` ns, but this is bad, especially in tests which will be looking for the NRT object on a specific ns (usually the one the tests components` are running in)

pkg/flagcodec/flagcodec.go
```diff
@@ -0,0 +1,105 @@
...
+ * Copyright 2021 Red Hat, Inc.
+ */
+
+// flagcodec allows to manipulate foreign command lines following the
+// standard golang conventions. It offeres two-way processing aka
+// parsing/marshalling of flags. It is different from the other, well
+// established packages (pflag...) because it aims to manipulate command
+// lines in general, not this program command line.
+
+package flagcodec
```
Tal-or, line 23: Cool! very nice package to do the trick

pkg/manifests/updates.go
```diff
@@ -100,25 +96,22 @@ func UpdateSchedulerConfigNamespaces(logger tlog.Logger, cm *corev1.ConfigMap, N
...
 	if len(ds.Spec.Template.Spec.Containers) > 1 {
 		// TODO: more polite/proper iteration
 		ds.Spec.Template.Spec.Containers[1].ImagePullPolicy = pullPolicy(pullIfNotPresent)
 	}
-	vars := map[string]string{
-		"RTE_POLL_INTERVAL": "10s",
-		"EXPORT_NAMESPACE":  ds.Namespace,
-	}
-	ds.Spec.Template.Spec.Containers[0].Command = UpdateResourceTopologyExporterCommand(ds.Spec.Template.Spec.Containers[0].Command, vars, plat)
+
```
cynepco3hahue, line 112: Can we move part of the command line configuration under `GetManifests` in general I am expecting that `UpdateResourceTopologyExporterDaemonSet` should update the only dynamical configuration? If each time it will apply the same arguments we definitely can move it under the `GetManifests` underlying methods.
Same true regarding `plat` parameter, like it no taste to pass it to the Update method when you should know the platform from the beginning and it should stay the same during the run.

pkg/manifests/yaml/rte/daemonset.yaml
fromanirh: good point, I should have been more explicit. Once https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/74 is merged, we will need no export namespace anymore. Actually the option will be gone entirely.
This means this PR should be merged only after https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/74 is merged.

pkg/manifests/updates.go
fromanirh: that's a good point. In general this is indeed the gist. There are a lot of nuances and poorly-specified areas, so let's gradually cleanup.

fromanirh: initial change: https://github.com/k8stopologyawareschedwg/deployer/pull/60/commits/e9ba6fa5d34329446d40e842ae3751efac60a8a3

we will probably have more in the future

fromanirh:
partially replaced by https://github.com/k8stopologyawareschedwg/deployer/pull/78

PR #61, opened 2021-11-16
Valid version in go.mod's require field

Tal-or:
Although this version being replaced, for some reason it still interfer when other components are trying to consume the deployer.
Changed the version into a valid one.


fromanirh:
I dunno exactly why this is needed, but will hardly hurt, so let's have it.

PR #63, opened 2021-11-17
sched: clusterrole: add missing permissions

fromanirh:
Add missing (intentionally RO) permissions to storage.k8s.io resources.
Looks like the core scheduler needs to watch these resources, and if it fails to watch them the scheduler doesn't initializes itself
correctly.

TL;DR: with the extra permissions, the scheduling works as expected;
without, it doesn't: pod requesting the secondary scheduler keep being Pending, even though NRT objects are populated correctly.

### DIFF
--- a/pkg/manifests/yaml/sched/scheduler/clusterrole.yaml
+++ b/pkg/manifests/yaml/sched/scheduler/clusterrole.yaml
@@ -54,7 +54,7 @@ rules:
   resources: ["subjectaccessreviews"]
   verbs: ["create"]
 - apiGroups: ["storage.k8s.io"]
-  resources: ["csinodes", "storageclasses"]
+  resources: ["csinodes", "storageclasses", "csidrivers", "csistoragecapacities"]
   verbs: ["get", "list", "watch"]
 - apiGroups: ["scheduling.sigs.k8s.io"]
   resources: ["podgroups", "elasticquotas"]


swatisehgal:
This is interesting, I was referring to https://github.com/kubernetes-sigs/scheduler-plugins/blob/6e58632f57f80274dc5e0f97df7f1d1907cc7c25/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml which is a reference point for RBACs needed for scheduler plugins. Don't see csidrivers, csistoragecapacities specified there. Might be worth submitting this this upstream as well?

fromanirh:
> This is interesting, I was referring to https://github.com/kubernetes-sigs/scheduler-plugins/blob/6e58632f57f80274dc5e0f97df7f1d1907cc7c25/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml which is a reference point for RBACs needed for scheduler plugins. Don't see csidrivers, csistoragecapacities specified there. Might be worth submitting this patch upstream as well?

surely we need to investigate more. Here we are running a pretty old snapshot, with this code we need these permissions, but it's possible they have actually been removed u/s.

fromanirh:
actually no, the file was introduced in that format and never changed

fromanirh:
looking at the permissions of the clusterrole `system:kube-scheduler` in kube 1.22.2 I see
```
- apiGroups:
  - storage.k8s.io
  resources:
  - csinodes
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - ""
  resources:
  - namespaces
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - storage.k8s.io
  resources:
  - csidrivers
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - storage.k8s.io
  resources:
  - csistoragecapacities
  verbs:
  - get
  - list
  - watch
```

Since (AFAIU) the secondary scheduler clusterrole wants to have the same permissions as this role, this explains why it is needed and yes, I will send a PR u/s

fromanirh:
https://github.com/kubernetes-sigs/scheduler-plugins/pull/294

swatisehgal:
Perfect, Thanks!

/lgtm

PR #64, opened 2021-11-22
Set static parameters during manifests load instead of updating it after

cynepco3hahue:

### DIFF
--- a/pkg/commands/render.go
+++ b/pkg/commands/render.go
@@ -86,7 +86,7 @@ func NewRenderSchedulerPluginCommand(commonOpts *CommonOptions, opts *renderOpti
 				return err
 			}
 
-			schedManifests, err := sched.GetManifests(commonOpts.UserPlatform)
+			schedManifests, err := sched.GetManifests(commonOpts.UserPlatform, rteNamespace)
 			if err != nil {
 				return err
 			}
@@ -129,7 +129,7 @@ func makeRTEObjects(commonOpts *CommonOptions) ([]client.Object, string, error) 
 		return nil, namespace, err
 	}
 
-	mf, err := rtemanifests.GetManifests(commonOpts.UserPlatform)
+	mf, err := rtemanifests.GetManifests(commonOpts.UserPlatform, namespace)
 	if err != nil {
 		return nil, namespace, err
 	}
@@ -161,7 +161,7 @@ func renderManifests(cmd *cobra.Command, commonOpts *CommonOptions, opts *render
 	}
 	objs = append(objs, rteObjs...)
 
-	schedManifests, err := sched.GetManifests(commonOpts.UserPlatform)
+	schedManifests, err := sched.GetManifests(commonOpts.UserPlatform, rteNs)
 	if err != nil {
 		return err
 	}
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -41,6 +41,8 @@ import (
 	apiconfig "sigs.k8s.io/scheduler-plugins/pkg/apis/config"
 
 	rteassets "github.com/k8stopologyawareschedwg/deployer/pkg/assets/rte"
+	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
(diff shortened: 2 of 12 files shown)


fromanirh:
there is a significant overlap with previous https://github.com/k8stopologyawareschedwg/deployer/pull/60
let's see how we can merge the two PRs.

cynepco3hahue:
> there is a significant overlap with previous #60
> let's see how we can merge the two PRs.

I do not have any problem first merging yours, just let not provide additional infra around arguments, because we should set arguments only once, at least now it is how it looks.

fromanirh:
> > there is a significant overlap with previous #60
> > let's see how we can merge the two PRs.
> 
> I do not have any problem first merging yours, just let not provide additional infra around arguments, because we should set arguments only once, at least now it is how it looks.

I agree my PR does much more (maybe too much for a single PR). So Let's move forward with yours, and I'll later rebase (/split) my PR on top. So I'll review and send the minimum amount of comments to move forward and merge.

cynepco3hahue:
> I agree my PR does much more (maybe too much for a single PR). So Let's move forward with yours, and I'll later rebase (/split) my PR on top. So I'll review and send the minimum amount of comments to move forward and merge.

Great, thanks!

pkg/commands/render.go
```diff
@@ -129,14 +129,16 @@ func makeRTEObjects(commonOpts *CommonOptions) ([]client.Object, string, error)
...
+	mf, err := rtemanifests.GetManifests(commonOpts.UserPlatform, namespace)
 	if err != nil {
 		return nil, namespace, err
 	}
 	mf = mf.Update(rtemanifests.UpdateOptions{
-		ConfigData:       commonOpts.RTEConfigData,
-		PullIfNotPresent: commonOpts.PullIfNotPresent,
-		Namespace:        namespace,
+		ConfigData: commonOpts.RTEConfigData,
+		DaemonSetOptions: &manifests.DaemonSetUpdateOptions{
```
fromanirh, line 138: I'm really not sold regarding the need of a substruct. Maybe we can move back `PullIfNotPresent` in the `rtemanifests.UpdateOptions`.

pkg/manifests/manifests.go
```diff
@@ -259,6 +288,42 @@ func DaemonSet(component string) (*appsv1.DaemonSet, error) {
 	if !ok {
 		return nil, fmt.Errorf("unexpected type, got %t", obj)
 	}
+
+	for i := range ds.Spec.Template.Spec.Containers {
+		c := &ds.Spec.Template.Spec.Containers[i]
+		if c.Name == containerNameRTE {
```
fromanirh, line 294: the other way around please:
```
if c.Name == containerNameRTE {
  continue
}
```

pkg/manifests/manifests.go
```diff
@@ -259,6 +288,42 @@ func DaemonSet(component string) (*appsv1.DaemonSet, error) {
...
+			c.Image = images.ResourceTopologyExporterImage
+			c.Command = []string{
```
fromanirh, line 296: this is better than before but I'm not really sold about it. But we can surely try and evolve incrementally.

fromanirh:
TL;DR: I'd avoid the `DaemonSetUpdateOptions` unless it's really really needed (let's do a deep cleanup later, since this is not going to be the final cleanup) and please let's simplify a bit the  update logic in https://github.com/k8stopologyawareschedwg/deployer/pull/64/files/0c9c92634445b37f0f8e6a333e0fe447fa00dec6#diff-1f9d2d18d4cabd8262e585b932330183f7cb9b544ac9f27c8bcb3ef9c8d14376 by swapping the guard. Everything else is good enough and we will improve incrementally.

pkg/commands/render.go
cynepco3hahue: yee, I thought I will need additional parameters. Will revert it

pkg/manifests/manifests.go
fromanirh: So let's go with this, we will improve later if needed once we have a clearer picture

fromanirh:
/lgtm
good enough simplification, we can incrementally improve later

PR #65, opened 2021-11-23
rte: fix permissions issues related to latest RTE changes

cynepco3hahue:
- grant NRT permissions on the cluster scope level
- grant permissions to update pod conditions under the RTE daemon set
namespace

### DIFF
--- a/pkg/manifests/rte/rte.go
+++ b/pkg/manifests/rte/rte.go
@@ -33,11 +33,13 @@ import (
 )
 
 type Manifests struct {
-	ServiceAccount *corev1.ServiceAccount
-	Role           *rbacv1.Role
-	RoleBinding    *rbacv1.RoleBinding
-	ConfigMap      *corev1.ConfigMap
-	DaemonSet      *appsv1.DaemonSet
+	ServiceAccount     *corev1.ServiceAccount
+	Role               *rbacv1.Role
+	RoleBinding        *rbacv1.RoleBinding
+	ClusterRole        *rbacv1.ClusterRole
+	ClusterRoleBinding *rbacv1.ClusterRoleBinding
+	ConfigMap          *corev1.ConfigMap
+	DaemonSet          *appsv1.DaemonSet
 
 	// OpenShift related components
 	MachineConfig             *machineconfigv1.MachineConfig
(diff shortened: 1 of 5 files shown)


Tal-or:
It lgtm but i'll leave @fromanirh the final approval

pkg/manifests/rte/rte.go
```diff
@@ -183,6 +192,8 @@ func (mf Manifests) ToCreatableObjects(hp *deployer.Helper, log tlog.Logger) []d
 	return append(objs,
 		deployer.WaitableObject{Obj: mf.Role},
 		deployer.WaitableObject{Obj: mf.RoleBinding},
+		deployer.WaitableObject{Obj: mf.ClusterRole},
+		deployer.WaitableObject{Obj: mf.ClusterRoleBinding},
```
fromanirh, line 196: when we transition to RTE >= v0.3.0 we will need to get rid of the role and rolebinding
We are not there yet

pkg/manifests/yaml/rte/role.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
...
 rules:
-- apiGroups: ["topology.node.k8s.io"]
-  resources: ["noderesourcetopologies"]
-  verbs: ["create", "update", "get", "list"]
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get"]
+- apiGroups: [""]
+  resources: ["pods/status"]
+  verbs: ["update"]
```
fromanirh, line 12: we need this to support RTE >= v0.3.0, but we are not there yet.
It won't hurt, and the support should be coming soon, so this time I don't mind running ahead of ourselves.

fromanirh (commented):
considering we want to support the API v0.0.12 and all the related components (RTE >= v0.3.0) and considering this PR only adds code and doesn't break the current flow, I think we can merge it.
We will cleanup later when we complete the transition to the v0.0.12 stack

pkg/manifests/rte/rte.go
cynepco3hahue: Why do we need to get rid of the role and role binding, otherwise I think we grant broad permissions to get pod and update pod status under the RTE manifest https://github.com/k8stopologyawareschedwg/resource-topology-exporter/blob/e57792df4a3b19928dbedf9cb23684f05acc1397/manifests/resource-topology-exporter-rbac.yaml#L15.
We should only have permissions to update daemon set pods under the specific namespace we do not need cluster scope role for it.

fromanirh: right, good point. I guess we will need them both after all.

PR #66, opened 2021-11-24
Update RTE and API versions

cynepco3hahue:
- advance RTE version to 0.3.1
- advance noderesourcetopology-api to v0.0.12


fromanirh:
nope, we cannot avance partially. We need to advance only when we can also bump the scheduler-plugin.

cynepco3hahue:
> nope, we cannot avance partially. We need to advance only when we can also bump the scheduler-plugin.

I am pretty dependent on it for the notify file.

cynepco3hahue:
@fromanirh What stop us to advance scheduler as well?

cynepco3hahue:
@fromanirh And some additional questions, do we have a table that correlates the RTE version with the scheduler plugin version?

fromanirh:
> > nope, we cannot avance partially. We need to advance only when we can also bump the scheduler-plugin.
> 
> I am pretty dependent on it for the notify file.

s

fromanirh:
sorry, wrong button :\

> @fromanirh What does stop us to advance scheduler as well?

The PR updating the scheduler plugin to NRT API v0.0.12 is not merged yet, AND we don't have a container image with that change in.


> @fromanirh And some additional questions, do we have a table that correlates the RTE version with the scheduler plugin version?

Nope, but we should add one in this repo

fromanirh:
issue about adding the compatibility matrix: https://github.com/k8stopologyawareschedwg/deployer/issues/67

PR #69, opened 2021-11-25
Always add notify file for both OpenShift and Kubernetes platform

cynepco3hahue:
Adding the notifier file will not harm under Kubernetes platform
and will make testing easier.


fromanirh (commented):
this seems fine, but don't we also need to add the parameter in the daemonset commandline?

cynepco3hahue:
> this seems fine, but don't we also need to add the parameter in the daemonset commandline?

True!

fromanirh (commented):
/approve
/lgtm
thanks for the quick fix!

PR #71, opened 2021-11-30
Mount configmap with the correct name

Tal-or:
We're creating the ConfigMap object with a name which is based on the DaemonSet's name,
but when we're mounting the ConfigMap as a volume we're using a constant name which is not the real ConfigMap's name.

The result is that the pod is getting deployed without a mounted ConfigMap.

### DIFF
--- a/pkg/manifests/rte/rte.go
+++ b/pkg/manifests/rte/rte.go
@@ -113,8 +113,12 @@ func (mf Manifests) Update(options UpdateOptions) Manifests {
 	ret.DaemonSet.Spec.Template.Spec.ServiceAccountName = mf.ServiceAccount.Name
 
 	rteConfigMapName := ""
+	if len(options.ConfigData) > 0 {
+		ret.ConfigMap = CreateConfigMap(ret.DaemonSet.Namespace, manifests.RTEConfigMapName, options.ConfigData)
+	}
+
 	if ret.ConfigMap != nil {
-		rteConfigMapName = manifests.RTEConfigMapName
+		rteConfigMapName = ret.ConfigMap.Name
 	}
 	manifests.UpdateResourceTopologyExporterDaemonSet(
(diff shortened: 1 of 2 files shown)


pkg/manifests/rte/rte.go
```diff
@@ -113,9 +113,14 @@ func (mf Manifests) Update(options UpdateOptions) Manifests {
 	ret.DaemonSet.Spec.Template.Spec.ServiceAccountName = mf.ServiceAccount.Name
 
 	rteConfigMapName := ""
+	if len(options.ConfigData) > 0 {
+		ret.ConfigMap = CreateConfigMap(ret.DaemonSet.Namespace, manifests.RTEConfigMapName, options.ConfigData)
+	}
+
 	if ret.ConfigMap != nil {
 		rteConfigMapName = manifests.RTEConfigMapName
```
fromanirh, line 121: we should probably do
```
        rteConfigMapName := ""
        if ret.ConfigMap != nil {
               rteConfigMapName = ret.ConfigMap.Name
        }
```

pkg/manifests/rte/rte.go
```diff
@@ -113,9 +113,14 @@ func (mf Manifests) Update(options UpdateOptions) Manifests {
...
 	}
+
```
fromanirh, line 123: if you re-upload, please remove unnecessary changes

pkg/manifests/updates.go
```diff
@@ -16,7 +16,7 @@ import (
 
 const (
 	metricsPort        = 2112
-	rteConfigMountName = "rte-config"
+	rteConfigMountName = "rte-config-volume"
```
fromanirh, line 19: is this still needed?

fromanirh (commented):
good catch! thanks for the PR!

pkg/manifests/updates.go
fromanirh: ok!

fromanirh (commented):
/lgtm

PR #72, opened 2021-11-30
Create ns for RTE on all platforms

Tal-or:

### DIFF
--- a/pkg/commands/render.go
+++ b/pkg/commands/render.go
@@ -139,10 +139,7 @@ func makeRTEObjects(commonOpts *CommonOptions) ([]client.Object, string, error) 
 	})
 
 	rteObjs := mf.ToObjects()
-	if commonOpts.UserPlatform == platform.Kubernetes {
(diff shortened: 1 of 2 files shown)


fromanirh:
now that we don't abuse the `openshift-monitoring` namespace anymore, this seems the right fix.

Tal-or:
I'm not fully aware if there is a historical reason for creating Namespace only under K8S, but definitely when using the deployer directly (not via numaresource-operator) the ns is needed.

fromanirh:
> I'm not fully aware if there is a historical reason for creating Namespace only under K8S, but definitely when using the deployer directly (not via numaresource-operator) the ns is needed.

Initially, I was piggybacking on `openshift-monitoring` namespace and on the `node-exporter` SCC to save time and to have a good-enough(ish) solution. Nowadays, thanks to the work of @cynepco3hahue , RTE goes cleanly with its own SCC, namespace and everything else.

PR #75, opened 2021-12-09
Delete scheduler rolebinding resource

Tal-or:
The rolebinding deployed at the kube-system ns and not in the scheduler ns, thus we need to delete it explicitly


fromanirh:
I seem to recall I had no specific reason to NOT delete this resource, so the fix here seems correct.

PR #76, opened 2021-12-12
Add podreadiness under daemonset

Tal-or:
This patch will allow RTE to use the podreadiness feature while we deploy it using the deployer/numaresource-operator

### DIFF
--- a/pkg/manifests/yaml/rte/daemonset.yaml
+++ b/pkg/manifests/yaml/rte/daemonset.yaml
@@ -11,6 +11,9 @@ spec:
       labels:
         name: resource-topology
     spec:
+      readinessGates:
+      - conditionType: "PodresourcesFetched"
+      - conditionType: "NodeTopologyUpdated"
       serviceAccountName: rte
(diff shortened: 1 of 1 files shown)


cynepco3hahue:
hm why do we need it? IIUC the controller should update it anyway, no?

Tal-or:
> hm why do we need it? IIUC the controller should update it anyway, no?

I looked at the controller code, I don't see any mention for that, I also looked at the RTE pod created by the controller, no mention for that either.

Besides IMHO it's a good practice to have it configured at the "source" (deployer) so even if the user will deploy RTE using the deployer directly (and not the controller) he'll have it.

cynepco3hahue:
I meant the RTE controller, that functionality that you added.

fromanirh:
since RTE supports this feature, I think it's useful to enable it.

Tal-or:
> I meant the RTE controller, that functionality that you added.

the RTE controller is numaresources-operator.
I didn't add any functionality there

PR #77, opened 2021-12-16
scheduler: allow to specify namespace

Tal-or:
This change is good in general, but also will help with implementing the scheduler deployment in NROP.
The change is completely optional and won't affect the default behavior

### DIFF
--- a/pkg/manifests/sched/sched.go
+++ b/pkg/manifests/sched/sched.go
@@ -79,6 +79,7 @@ func (mf Manifests) Clone() Manifests {
 type UpdateOptions struct {
 	Replicas         int32
 	PullIfNotPresent bool
+	Namespace        string
 }
 
 func (mf Manifests) Update(logger tlog.Logger, options UpdateOptions) Manifests {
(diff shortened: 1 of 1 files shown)


fromanirh:
I don't see the benefit for the deployer here, could you please elaborate?

Tal-or:
> I don't see the benefit for the deployer here, could you please elaborate?

It'll allow to select the namespace where you want the scheduler to get deployed

fromanirh:
> > I don't see the benefit for the deployer here, could you please elaborate?
> 
> It'll allow to select the namespace where you want the scheduler to get deployed

but in the deployer we do want to have opinionated defaults. Selecting the namespace will create more complications and we don't have a usecase for that yet.

So maybe let's start with the use case in the contex to of the deployer, so we can evaluate the best solution.

Tal-or:
> but in the deployer we do want to have opinionated defaults.

And we do, in case that the namespace is not provided we're staying with the default. 
Changing the namespace is totally optional

> So maybe let's start with the use case in the contex to of the deployer, so we can evaluate the best solution.
I mentioned the context in the PR, but I can elaborate more if needed

Anyhow, I also have a workaround in case we don't want to have this patch in the deployer (it's just that the workaround is a bit ugly and has some copy-paste from the original deployer code)

fromanirh:
> > but in the deployer we do want to have opinionated defaults.
> 
> And we do, in case that the namespace is not provided we're staying with the default. Changing the namespace is totally optional

True, but then we need to support (and test) this option, whose benefit is still not clear to me

fromanirh:
wrong button

Tal-or:
> True, but then we need to support (and test) this option, whose benefit is still not clear to me

OK let's leave it for now, we can discuss this PR again once I'll have the changes in NROP ready.
Then I believe the benefit will be more clear.

PR #79, opened 2022-01-03
manifests: rte: split args from command

fromanirh:
Per the documentation:
```
When you create a Pod, you can define a command and arguments for the containers that run in the Pod.
To define a command, include the `command` field in the configuration file. To define arguments for
the command, include the `args` field in the configuration file.
The command and arguments that you define cannot be changed after the Pod is created.
```
(see:
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/
see also:
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes)

The most correct and precise form is thus to supply *both* `command` and
`args` - and that was the original intent of the manifest; conveying
everything into the `command` is a slightly abuse, even if this works.

### DIFF
--- a/pkg/manifests/manifests.go
+++ b/pkg/manifests/manifests.go
@@ -373,8 +373,11 @@ func DaemonSet(component string, plat platform.Platform, namespace string) (*app
 		c := &ds.Spec.Template.Spec.Containers[i]
 		if c.Name == containerNameRTE {
 			c.Image = images.ResourceTopologyExporterImage
+			// we do this explicitely, but should be already OK from the YAML manifest
 			c.Command = []string{
 				"/bin/resource-topology-exporter",
+			}
+			c.Args = []string{
 				"--sleep-interval=10s",
 				fmt.Sprintf("--sysfs=%s", containerHostSysDir),
 				fmt.Sprintf("--podresources-socket=unix://%s", containerPodResourcesSocket),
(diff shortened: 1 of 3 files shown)


Tal-or:
Changes also needed at:
https://github.com/k8stopologyawareschedwg/deployer/blob/0db4e73481514d40439e07f6f34dc57e88f85dd8/pkg/manifests/manifests.go#L376
and
https://github.com/k8stopologyawareschedwg/deployer/blob/0db4e73481514d40439e07f6f34dc57e88f85dd8/pkg/manifests/manifests.go#L403

fromanirh:
> Changes also needed at:
> 
> https://github.com/k8stopologyawareschedwg/deployer/blob/0db4e73481514d40439e07f6f34dc57e88f85dd8/pkg/manifests/manifests.go#L376
> 
> and
> 
> https://github.com/k8stopologyawareschedwg/deployer/blob/0db4e73481514d40439e07f6f34dc57e88f85dd8/pkg/manifests/manifests.go#L403

thanks, fixed

fromanirh:
also tested locally, the process command line looks as expected
```bash
K exec -n tas-topology-updater resource-topology-exporter-mc668 -c resource-topology-exporter -- /bin/sh
K exec -ti -n tas-topology-updater resource-topology-exporter-mc668 -c resource-topology-exporter -- /bin/sh
sh-4.2# ps -fauxw
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root          68  0.0  0.0  11840  2836 pts/0    Ss   09:29   0:00 /bin/sh
root          75  0.0  0.0  51744  3504 pts/0    R+   09:29   0:00  \_ ps -fauxw
root           1  0.3  0.0 740752 32504 ?        Ssl  09:27   0:00 /bin/resource-topology-exporter --sleep-interval=10s --sysfs=/host-sys --podresources-socket=unix:///host-podresources-socket/kubelet.sock --notify-file=/host-run-rte/notify --kubelet-config-file=/host-var-lib-kubelet/config.yaml --kubelet-state-dir=/
sh-4.2# exit
```

PR #83, opened 2022-02-13
validate: extend to check and store the cluster version

fromanirh:
Overhaul the `validate` package to check and store the cluster version.
Breaks the API - will require a major bump.
    
In this change, we add support to check the cluster version, enabling smarter validation check down the road (e.g. feature gates).
Now validation helpers can (and should) alter the validation state accumulating intermediate results - while still returning their specific validation results.
    
Same goes for the cluster version - it needs to be stored in the Validator object, and possibly to be checked first. This is why    we introduce and recommend create functions.

### DIFF
--- a/pkg/validator/kubeletconfig_test.go
+++ b/pkg/validator/kubeletconfig_test.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/version"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
 )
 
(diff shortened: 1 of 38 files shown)


pkg/validator/kubeletconfig_test.go
```diff
@@ -201,7 +266,7 @@ func TestKubeletValidations(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			got := vd.ValidateNodeKubeletConfig(nodeName, tc.kubeletConf)
+			got := vd.ValidateNodeKubeletConfig(nodeName, nil, tc.kubeletConf)
```
cynepco3hahue, line 269: I am missing some test cases with the specific k8s version, =1.23 and < 1.23

fromanirh: yeah, I'll add them.

cynepco3hahue (approved):

PR #92, opened 2022-03-27
vendor: bump scheduler-plugins to v0.22.6 

Tal-or:
Bump to latest scheduler-plugins release so we could use  ```KubeSchedulerConfig v1beta2``` API version


go.mod
```diff
@@ -99,3 +99,5 @@ replace (
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20211105154855-cb9596dd5fba // release-4.10
 	github.com/openshift/machine-config-operator => github.com/openshift/machine-config-operator v0.0.1-0.20211105081319-76d6155c1dab // release-4.10
 )
+
+replace sigs.k8s.io/scheduler-plugins => sigs.k8s.io/scheduler-plugins v0.22.6
```
fromanirh, line 103: before using replace, I'd start bumping `scheduler-plugins` in the top-level `require` block

fromanirh (commented):
yes, is time this projects starts moving forward with the scheduler-plugin deps. Please check the inline comment

go.mod
Tal-or: I thought that using replace is a best practice convention, but sure i'll change it

fromanirh (approved):
/lgtm

PR #96, opened 2022-03-28
api:vendor: move deployer to use v1beta2's scheduler-plugins scheme

Tal-or:
For the go-client and controller runtime to be able to work with the newer version of KubeSchedulerConfig type, we must use add the correct version to the scheme.


fromanirh:
I'm not sure that's everything we need, but this seems to be heading in the right direction

PR #105, opened 2022-07-12
images: RTE: bump to v0.6.0

fromanirh:
Previously, we just neglected to update the builtin images, trusting the user to replace them. Time to fix the defaults
to something more modern than ancient 0.3.1, but still compatible with the NRT API 0.0.12.
The scheduler side will be updated with a different PR.


swatisehgal:
Let's update the scheduler images as well (sounds good to do it in a different PR).

/lgtm

fromanirh:
> Let's update the scheduler images as well (sounds good to do it in a different PR).
> 
> /lgtm

indeed. We will need to have regular builds however, or to revert to u/s builds.

PR #107, opened 2022-07-13
selinuxpolicy: updates to enable support for OCP 4.11+/K8s 1.24+

swatisehgal:
In container-selinux versions v2.186.0, v2.187.0 and v2.188.0, due to the updates here:
https://github.com/containers/container-selinux/commit/cf704e467c9b421f0c803f437b716e0b0315c1b6
selinuxpolicy updates are required (updating `container_runtime_t` to `kubelet_t`).

This is consumed in OCP 4.11 and K8s 1.24 and we need to make sure that the selinuxpolicy is
appropriately updated.

In addition to that, we need to continue supporting the older selinuxpolicy supplied in older
versions (e.g. 4.10) and that is the reason this patch is proposing an additional selinuxpolicy
and not completely replacing the original one.

Signed-off-by: Swati Sehgal <swsehgal@redhat.com>


pkg/assets/rte/assets.go
```diff
@@ -10,6 +10,9 @@ var SELinuxInstallSystemdServiceTemplate []byte
 //go:embed selinuxpolicy.cil
 var SELinuxPolicy []byte
```
fromanirh, line 11: let's rename this `SELinuxPolicy410` - we're breaking compat already and we will release 0.7.0

pkg/assets/rte/assets.go
```diff
@@ -10,6 +10,9 @@ var SELinuxInstallSystemdServiceTemplate []byte
...
 
+//go:embed selinuxpolicy-latest.cil
+var SELinuxPolicyLatest []byte
```
fromanirh, line 14: `Latest` is a floating tag :) Let's call it `SELinuxPolicyOCP411`

pkg/assets/rte/selinuxpolicy-latest.cil
```diff
@@ -0,0 +1,21 @@
...
+
+	;
+	; Allow to RTE pod access to /run/rte directory
+	(allow process container_var_run_t (dir (add_name write)))
+	(allow process container_var_run_t (file (create read write open)))
+
+	;
+	; Allow to RTE pod connect, read and write permissions to /var/lib/kubelet/pod-resource/kubelet.sock
+	(allow process container_var_lib_t (sock_file (open getattr read write ioctl lock append)))
+	(allow process kubelet_t (unix_stream_socket (connectto)))
```
fromanirh, line 20: this seems correct - bar it should be renamed `selinuxpolicy.ocp411.cil` (or -ocp411.cil or whatever including ocp and 411)

pkg/assets/rte/assets.go
```diff
@@ -10,6 +10,9 @@ var SELinuxInstallSystemdServiceTemplate []byte
 //go:embed selinuxpolicy.cil
```
fromanirh, line 10: let's rename this `selinuxpolicy.ocp4.10.cil` (or whatever, see below)

pkg/manifests/manifests.go
```diff
@@ -82,6 +83,11 @@ const (
 	hostNotifierDir                 = "/run/rte"
 )
 
+const (
+	latesK8sVersion  = "1.24"
```
fromanirh, line 87: the problem here is "latest" is "latest at time of writing", which will be obsolete pretty fast. Let's be explicit with versions

pkg/manifests/manifests.go
```diff
@@ -82,6 +83,11 @@ const (
...
+	latestOCPVersion = "4.11"
+)
+
```
fromanirh, line 90: maybe let's add a function - which is usually better than exposing plain vars:
```
func GetSELinuxPolicy(ver platform.Version) []byte {
       // error should never happen: we control the input here
       if ok, _ := ver.AtLeastString("4.11"); ok {
               return SELinuxPolicyOCP411
       }
       return SELinuxPolicyOCP410
}
```

pkg/manifests/manifests.go
```diff
@@ -455,9 +461,41 @@ func MachineConfig(component string) (*machineconfigv1.MachineConfig, error) {
 
 func getIgnitionConfig() ([]byte, error) {
```
fromanirh, line 462: the core idea is correct, but this logic should not be done implicitely - and better not make calls querying the cluster state in this code. Let's just pass it the desired version as flag and let's get and pass the flag in the upper layers.

what about
```
func getInitionConfig(ver platform.Version) ([]byte, error) {
       selinuixPolicy := rteassets.GetSELinuxPolicy(ver)
       ...
}
```

fromanirh:
very nice start!

pkg/assets/rte/assets.go
swatisehgal: Yep, makes sense.

pkg/assets/rte/assets.go
swatisehgal: Ack, will update now.

pkg/manifests/manifests.go
swatisehgal: Yeah, that will certainly make the code nice and clean. Thanks for the input.

pkg/manifests/rte/rte.go
```diff
@@ -250,7 +250,7 @@ func GetManifests(plat platform.Platform, namespace string) (Manifests, error) {
 	mf := New(plat)
 
 	if plat == platform.OpenShift {
-		mf.MachineConfig, err = manifests.MachineConfig(manifests.ComponentResourceTopologyExporter)
+		mf.MachineConfig, err = manifests.MachineConfig(manifests.ComponentResourceTopologyExporter, platform.Version(plat))
```
fromanirh, line 253: I think you want to use `detect.Version(plat)`  here?

pkg/assets/rte/assets.go
```diff
@@ -2,16 +2,33 @@ package rte
...
+//go:embed selinuxpolicy-ocp411.cil
+var SELinuxPolicyOCP411 []byte
 
 //go:embed hookconfigrtenotifier.json.template
 var HookConfigRTENotifier []byte
 
 //go:embed rte-notifier.sh
 var NotifierScript []byte
+
+func GetSELinuxPolicy(ver platform.Version) []byte {
```
fromanirh, line 28: if I'm right (see below), it seems this must return `([]byte, error)`

pkg/manifests/rte/rte.go
swatisehgal: Ah yeah, you are right! Fixing.

pkg/assets/rte/assets.go
swatisehgal: I am not sure I understand, we are returning either `SELinuxPolicyOCP411` or `SELinuxPolicyOCP412`  which are both of type []byte. Why you we need to return `([]byte, error)` ?

fromanirh: because this function should have failed when fed with `plat` casted as `platform.Version`, and we swallow the error, so the test didn't fail, which is bad

swatisehgal: Fixing it now.

fromanirh: yes, it's on me, sorry for misleading you. I was too optimist!

fromanirh:
thanks for the updates. Looks very good. It seems to me that we only miss to update the `pkg/commands/deploy.go` source/command (we're running out of good names!), like this:
```
diff --git a/pkg/commands/deploy.go b/pkg/commands/deploy.go
index 4bdcba0..8d24a7c 100644
--- a/pkg/commands/deploy.go
+++ b/pkg/commands/deploy.go
@@ -30,6 +30,7 @@ import (
 
 type DeployOptions struct {
        clusterPlatform platform.Platform
+       clusterVersion  platform.Version
        waitCompletion  bool
 }
 
@@ -62,6 +63,11 @@ func NewRemoveCommand(commonOpts *CommonOptions) *cobra.Command {
                        if opts.clusterPlatform == platform.Unknown {
                                return fmt.Errorf("cannot autodetect the platform, and no platform given")
                        }
+                       verDetect := detectVersion(commonOpts.DebugLog, platDetect.Discovered, commonOpts.UserPlatformVersion)
+                       opts.clusterVersion = verDetect.Discovered
+                       if opts.clusterVersion == platform.MissingVersion {
+                               return fmt.Errorf("cannot autodetect the platform version, and no version given")
+                       }
```

similar changes would be needed all across `pkg/commands/deploy.go` if I'm right
Unfortunately it's likely the existing e2e test don't cover well enough this flow. If nothing else, we're still on 1.23 in our CI lane.
Surely improving CI can be deferred to another PR.

test/e2e/positive.go
```diff
@@ -215,7 +215,7 @@ var _ = ginkgo.Describe("[PositiveFlow] Deployer execution", func() {
 				ns, err := manifests.Namespace(manifests.ComponentResourceTopologyExporter)
 				gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-				mf, err := rte.GetManifests(platform.Kubernetes, ns.Name)
+				mf, err := rte.GetManifests(platform.Kubernetes, platform.Version("1.23"), ns.Name)
```
fromanirh, line 218: please squash this commit in the previous

pkg/assets/rte/assets.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 const (
-	OCPVersion411 = "4.11"
+	OCPVersion411 = "v4.11"
```
fromanirh, line 10: could you please elaborate in the commit message why this is needed? If this makes some test fail, let's try to squash this commit in the one introducing the change (which then broke)

fromanirh (commented):
minor improvements inside

pkg/assets/rte/assets.go
swatisehgal: We are using [ParseVersion](https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/commands/root.go#L121) to parse the version which inturn calls [goversion.Parse](https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/deployer/platform/version.go#L26) that checks if the version string matches the [regex](https://github.com/k8stopologyawareschedwg/deployer/blob/main/vendor/github.com/aquasecurity/go-version/pkg/version/version.go#L22-L25). Therefore we need to prefix version with the letter `v`. Also I noticed that in the example where we ask for platform explicitly we have an [example](https://github.com/k8stopologyawareschedwg/deployer/blob/main/pkg/commands/root.go#L93) that satisfies above.

swatisehgal: I think it is best to squash this commit as well.

swatisehgal:
Need to tidy up a couple of things, on it now.

pkg/commands/deploy.go
```diff
@@ -64,6 +65,12 @@ func NewRemoveCommand(commonOpts *CommonOptions) *cobra.Command {
 			if opts.clusterPlatform == platform.Unknown {
 				return fmt.Errorf("cannot autodetect the platform, and no platform given")
 			}
+			versionDetect, source := detect.FindVersion(platDetect.Discovered, commonOpts.UserPlatformVersion)
+			commonOpts.DebugLog.Printf("Version detection source: %s", source)
+			opts.clusterVersion = versionDetect.Discovered
+			if opts.clusterVersion == platform.MissingVersion {
+				return fmt.Errorf("cannot autodetect the platform version, and no version given")
+			}
```
fromanirh, line 73: please pass `Version: opts.ClusterVersion` down when we call `updaters.Remove` (you may need to change `updaters.Options`)

pkg/commands/deploy.go
```diff
@@ -162,6 +181,12 @@ func NewDeployTopologyUpdaterCommand(commonOpts *CommonOptions, opts *DeployOpti
...
```
fromanirh, line 189: please pass `Version: opts.ClusterVersion` down when we call `updaters.Deploy` (you may need to change `updaters.Options`)

pkg/commands/deploy.go
```diff
@@ -233,6 +269,12 @@ func NewRemoveTopologyUpdaterCommand(commonOpts *CommonOptions, opts *DeployOpti
...
```
fromanirh, line 277: please pass `Version: opts.ClusterVersion` down when we call `updaters.Remove` (you may need to change `updaters.Options`)

pkg/commands/deploy.go
```diff
@@ -253,6 +295,12 @@ func deployOnCluster(commonOpts *CommonOptions, opts *DeployOptions) error {
 	if opts.clusterPlatform == platform.Unknown {
 		return fmt.Errorf("cannot autodetect the platform, and no platform given")
 	}
+	versionDetect, source := detect.FindVersion(platDetect.Discovered, commonOpts.UserPlatformVersion)
+	commonOpts.DebugLog.Printf("Version detection source: %s", source)
+	opts.clusterVersion = versionDetect.Discovered
+	if opts.clusterVersion == platform.MissingVersion {
+		return fmt.Errorf("cannot autodetect the platform version, and no version given")
+	}
```
fromanirh, line 303: please pass `Version: opts.ClusterVersion` down when we call `updaters.Deploy` (you may need to change `updaters.Options`)

fromanirh (commented):
almost there. You properly detect the version in the right places, but we need to pass it through another layer

pkg/commands/deploy.go
swatisehgal: Ah, sorry about that. Will fix now.

swatisehgal: I had already updated the `Updaters.Options` but missed assigning version to it.

fromanirh: > I had already updated the `Updaters.Options` but missed assigning version to it.

right, the change is fine as you did. Just pass the value and we should be done.

PR #116, opened 2022-09-08
manifests: ocp: rte: remove hardcoded TM setting

fromanirh:
Long, long ago, we added a OCP-specific setting to fix the TM policy, because OCP doesn't allow to read the kubelet config.

Since then, OCP made progress and provided better option to tell RTE which is the TM policy configured (e.g. operator).

We can and should now get rid of this hardcoded default.


swatisehgal (approved):
/lgtm

PR #148, opened 2023-03-03
context: add and use explicit context in libraries

ffromani:
Up until now, the project libraries hardcoded `context.TODO()` instead of requiring a context from the caller.
This is a bad practice and makes impossible to cancel the APIs, even when the underlyingl libraries allowed to in turn accepting a context.

Fix the API to accept contexts, and pass them through as needed.

### DIFF
--- a/pkg/deployer/wait/daemonset.go
+++ b/pkg/deployer/wait/daemonset.go
@@ -27,10 +27,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func ForDaemonSetReadyByKey(cli client.Client, logger logr.Logger, key ObjectKey, pollInterval, pollTimeout time.Duration) (*appsv1.DaemonSet, error) {
+func ForDaemonSetReadyByKey(ctx context.Context, cli client.Client, logger logr.Logger, key ObjectKey, pollInterval, pollTimeout time.Duration) (*appsv1.DaemonSet, error) {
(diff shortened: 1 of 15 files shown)


ffromani:
Fixes: #149

pkg/deployer/wait/daemonset.go
```diff
@@ -27,10 +27,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func ForDaemonSetReadyByKey(cli client.Client, logger logr.Logger, key ObjectKey, pollInterval, pollTimeout time.Duration) (*appsv1.DaemonSet, error) {
+func ForDaemonSetReadyByKey(cli client.Client, ctx context.Context, logger logr.Logger, key ObjectKey, pollInterval, pollTimeout time.Duration) (*appsv1.DaemonSet, error) {
```
swatisehgal, line 30: I have noticed in Kubernetes codebase that `ctx context.Context` is usually the first parameter in a function definition. I guess it is a convention but it doesn't hurt to follow it :) Do you mind updating here and in a few other placed below so that `ctx context.Context` is passed as the first parameter?

ffromani: sure thing, I don't have a preference here so I updated the PR.

swatisehgal (approved):
/lgtm
/approve

Thanks for addressing my comment.

PR #167, opened 2023-03-22
selinux: update ocp policies

ffromani:
catch up with container-selinux v2.188 and 2.199


ffromani:
> Could you please add some references (for documentation purposes) about the meaning of the changes you did?

I expect more work in this area, I'll add comments when the definitive fix lands

Issue #177, opened 2023-04-04
manifests: scheduler should run in the control plane (default, but overridable)


ffromani:
tune the manifests to (optionally) make sure the scheduler do run in the control plane by default, but allow to override this option and behave like it is today (scheduler runs on worker nodes)

ffromani:
we will change defaults and defer the overridable part to later PRs

PR #199, opened 2023-06-14
sched: manifests: run on the control plane

ffromani:
the scheduler is a control plane component, thus it should run by default on the control plane.

Fixes: #177


ffromani:
HOLD! failures are legit and needs to be investigated. And e2e tests can use some improvements

ffromani:
failures fixed... by disabling the option by default on CI :\

PR #202, opened 2023-06-15
pkg: trickle down defaults in pkgs

ffromani:
Move defaults closer to the sub-packages instead
of declaring them in the root command and just imply consistency all across the stack.

Now we have means to programmatically access them.


ffromani:
needs deeper refactoring

PR #207, opened 2023-08-02
manifests: updater mount the directory, not the socket

ffromani:
Quoting kube docs:

```
when accessing the `/var/lib/kubelet/pod-resources/kubelet.sock` from DaemonSet
or any other app deployed as a container on the host, which is mounting socket as
a volume, it is a good practice to mount directory `/var/lib/kubelet/pod-resources/`
instead of the `/var/lib/kubelet/pod-resources/kubelet.sock`. This will ensure
that after kubelet restart, container will be able to re-connect to this socket.

Container mounts are managed by inode referencing the socket or directory,
depending on what was mounted. When kubelet restarts, socket is deleted
and a new socket is created, while directory stays untouched.
So the original inode for the socket become unusable. Inode to directory
will continue working.
```

### DIFF
--- a/pkg/manifests/yaml/nfd/topologyupdater/daemonset.yaml
+++ b/pkg/manifests/yaml/nfd/topologyupdater/daemonset.yaml
@@ -39,8 +39,8 @@ spec:
           volumeMounts:
             - mountPath: /host-var/lib/kubelet/config.yaml
               name: kubelet-podresources-conf
-            - mountPath: /host-var/lib/kubelet/pod-resources/kubelet.sock
-              name: kubelet-podresources-sock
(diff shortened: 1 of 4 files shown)


Tal-or:
This is interesting. So IOW mounting a volume as socket and restarting kubelet causes the container to failed to connect the socket?

ffromani:
> This is interesting. So IOW mounting a volume as socket and restarting kubelet causes the container to failed to connect the socket?

I never tested or seen this myself in the wild, but there are reports in the community of similar incidents with other software components mounting in a similar way. The rationale for the change seems solid and the change is expected to make no difference in the happy path, so I'm proposing the fix preemptively.

Issue #208, opened 2023-08-02
bump deps for v0.15.0


ffromani:
we should bump deps to
- scheduler plugins v0.26.z
- RTE v0.13 (still not out)
- NFD, possibly (needs to be checked)

PR #213, opened 2023-08-22
images: bump payload 20230822

ffromani:
bump payload images, and fixes the images command output along the way

Fixes: #208


ffromani:
turns out we found a bug in NFD. If we disable the state dirs monitoring, the timer-based monitoring is also disabled. This is not correct, the state dirs monitoring should augment the timer. Will followup ASAP.

ffromani:
> turns out we found a bug in NFD. If we disable the state dirs monitoring, the timer-based monitoring is also disabled. This is not correct, the state dirs monitoring should augment the timer. Will followup ASAP.

https://github.com/kubernetes-sigs/node-feature-discovery/pull/1325

PR #214, opened 2023-09-01
statedir: forcefully disable

ffromani:
The kubelet state dir monitoring is hard to secure and has limitations (cannot react timely to deletions).
The preferred approach is to use OCI hooks and notification file.
So we unconditionally disable on NFD and stop enabling it on RTE.


ffromani:
nope, has to be done in the same PR as #623

PR #216, opened 2023-09-14
wait: enhance functionality

jlojosnegros:
Update currently used wait utilities to conform with the same functions in NRO as a first step to merge both and avoid duplicates.

issue: https://github.com/openshift-kni/numaresources-operator/issues/530

### DIFF
--- a/pkg/deployer/wait/deployment.go
+++ b/pkg/deployer/wait/deployment.go
@@ -24,7 +24,16 @@ import (
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
 )
 
-func (wt Waiter) ForDeploymentCompleteByKey(ctx context.Context, key ObjectKey, replicas int32) (*appsv1.Deployment, error) {
+// This function waits for the readiness of the pods under the deployment. The best use of this check is for
+// completely new deployments. If the deployment exists on the cluster and simply updated, this check is
+// not enough to guarantee that the deployment is ready with the NEW replica, thus need to cover that by
+// additional checks as the context requires
+func (wt Waiter) ForDeploymentComplete(ctx context.Context, dp *appsv1.Deployment) (*appsv1.Deployment, error) {
+	if dp.Spec.Replicas == nil {
+		return nil, fmt.Errorf("unspecified replicas in %s/%s", dp.Namespace, dp.Name)
+	}
+
+	key := ObjectKeyFromObject(dp)
 	updatedDp := &appsv1.Deployment{}
 	err := k8swait.PollImmediate(wt.PollInterval, wt.PollTimeout, func() (bool, error) {
 		err := wt.Cli.Get(ctx, key.AsKey(), updatedDp)
@@ -33,7 +42,7 @@ func (wt Waiter) ForDeploymentCompleteByKey(ctx context.Context, key ObjectKey, 
 			return false, err
 		}
 
-		if !areDeploymentReplicasAvailable(&updatedDp.Status, replicas) {
+		if !isDeploymentComplete(dp, &updatedDp.Status) {
 			wt.Log.Info("deployment not complete",
 				"key", key.String(),
 				"replicas", updatedDp.Status.Replicas,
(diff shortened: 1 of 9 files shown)


pkg/deployer/wait/namespace.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+
+	k8swait "k8s.io/apimachinery/pkg/util/wait"
+)
+
+func (wt Waiter) ForNamespaceDeleted(ctx context.Context, namespace string) error {
```
ffromani, line 27: this is useful and should probably be used in `pkg/commands`

pkg/deployer/wait/wait.go
```diff
@@ -98,15 +98,9 @@ func (wt *Waiter) Interval(iv time.Duration) *Waiter {
 	return wt
 }
 
-func (wt Waiter) ForNamespaceDeleted(ctx context.Context, namespace string) error {
```
ffromani, line 101: why was this function deleted?

jlojosnegros: Not deleted but moved on its own file

ffromani: ack, thanks

pkg/deployer/wait/namespace.go
jlojosnegros: It was already been used in commands.

pkg/deployer/wait/wait.go
```diff
@@ -98,15 +98,9 @@ func (wt *Waiter) Interval(iv time.Duration) *Waiter {
...
-	log.Info("wait for the namespace to be gone")
-	return k8swait.PollImmediate(wt.PollInterval, wt.PollTimeout, func() (bool, error) {
-		nsKey := ObjectKey{Name: namespace}
-		ns := corev1.Namespace{} // unused
-		err := wt.Cli.Get(ctx, nsKey.AsKey(), &ns)
-		return deletionStatusFromError(wt.Log, "Namespace", nsKey, err)
-	})
+func (wt *Waiter) Steps(st int) *Waiter {
+	wt.PollSteps = st
+	return wt
```
ffromani, line 102: where is this used?

pkg/objectwait/rte/rte.go
```diff
@@ -94,9 +114,64 @@ func Deletable(mf rtemf.Manifests, cli client.Client, log logr.Logger) []objectw
...
+					return err
+				}
+				return waitForMachineConfigPoolsCondition(cli, log, ctx, mcps, mcov1.MachineConfigPoolUpdated, pollInterval, pollTimeout)
+			},
 		})
 	}
 	return objs
 }
+
+func waitForMachineConfigPoolsCondition(cli client.Client, log logr.Logger, ctx context.Context, mcps []*mcov1.MachineConfigPool, cond mcov1.MachineConfigPoolConditionType, interval, timeout time.Duration) error {
```
ffromani, line 136: this function could fit in the `wait` package, perhaps later.
The biggest problem however is we don't have OCP in our CI yet (and it's hard to get fixed) so we either add robust unit test coverage, we start using mock aggressively or we postpone the additions of more OCP-specific code which isn't quite trivial.

I'm fine with any of the above approaches.

pkg/deployer/wait/wait.go
jlojosnegros: You are right, this one escape to my deletion effort.

Thanks

pkg/objectwait/rte/rte.go
jlojosnegros: I would like to give it a try to the test coverage.
I am not sure though if there is a mock framework of preference

pkg/deployer/wait/deployment.go
```diff
@@ -24,7 +24,16 @@ import (
 	k8swait "k8s.io/apimachinery/pkg/util/wait"
 )
 
-func (wt Waiter) ForDeploymentCompleteByKey(ctx context.Context, key ObjectKey, replicas int32) (*appsv1.Deployment, error) {
+// This function waits for the readiness of the pods under the deployment. The best use of this check is for
```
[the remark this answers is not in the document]
ffromani: besides, why only this function has a docstring?

pkg/objectwait/rte/rte.go
```diff
@@ -94,9 +114,64 @@ func Deletable(mf rtemf.Manifests, cli client.Client, log logr.Logger) []objectw
...
+	for _, mcp := range mcps {
+		go func(mcp *mcov1.MachineConfigPool) {
+			err := wait.With(cli, log).
+				Interval(interval).
+				Timeout(timeout).
+				ForMachineConfigPoolCondition(ctx, mcp, cond)
+			c1 <- err
+		}(mcp)
+	}
+	var errs multierrors.MultiErrors
```
ffromani, line 147: it seems to me that we need at least 1 error to make the wait fail. IOW, it takes any error to fail. If so we should just return the first one got, while logging all of them. This can save us the addition of `multierror`

ffromani (commented):
are we planning to cover the OCP/MCP changes only with the newly added unit tests?

PR #225, opened 2023-10-22
scheduler: update: use unstructured in RenderConfig

ffromani:
Rewrite RenderConfig to use `unstructured` vs typed API.
The rationale for this change is that doing so will enable us to drop all the dependencies to the scheduler-plugins
API packages, which pull the scheduler API package, which eventually pulls in k8s.io/kubernetes, which we should
really NOT depend on.

We take the change to improve the RenderConfig API and make it forward extensible, anticipating the future support
for more scheduler plugin parameters.

The end result in terms of rendered YAML is expected to be not 1:1 indentical, but fully compatible.


ffromani:
/hold

looks promising, but needs more testing

pkg/objectupdate/sched/render.go
```diff
@@ -0,0 +1,165 @@
...
+	newData, err := RenderConfig([]byte(data), schedulerName, params)
+	if err != nil {
+		return err
+	}
+
+	cm.Data[SchedulerConfigFileName] = string(newData)
+	return nil
+}
+
+func RenderConfig(data []byte, schedulerName string, params *ConfigParams) ([]byte, error) {
```
Tal-or, line 61: Is this change means that in case of updates to the API (mainly addition of new fields) we would need to update this function as well?

PR #229, opened 2023-10-23
gh actions: run unit tests on all branches

ffromani:
these tests are cheap, let's do them.


ffromani:
backporting down to 0.13, included

PR #253, opened 2023-11-29
pkg: rte: address verbosity flag properly

shajmakh:
Follow the convention used in NROP repo to address container args with two "-".  -v is never actually updated while --v is. Keep the convention and use it like "--v".

### DIFF
--- a/pkg/objectupdate/rte/rte.go
+++ b/pkg/objectupdate/rte/rte.go
@@ -123,7 +123,7 @@ func DaemonSet(ds *appsv1.DaemonSet, plat platform.Platform, configMapName strin
 		}
 
 		flags := flagcodec.ParseArgvKeyValue(cntSpec.Args)
-		flags.SetOption("-v", fmt.Sprintf("%d", opts.Verbose))
+		flags.SetOption("--v", fmt.Sprintf("%d", opts.Verbose))
 		if opts.UpdateInterval > 0 {
 			flags.SetOption("--sleep-interval", fmt.Sprintf("%v", opts.UpdateInterval))
 		} else {


ffromani (commented):
thanks for raising this PR. Since this project is acting as frontend for others (RTE, NFD...) we need to document the conventions. I'm inclined to keep single-letter options with single dash (`-v`) because this is reminiscent of the unix convention.

Do we have a test which reproduces this issue?

shajmakh:
> thanks for raising this PR. Since this project is acting as frontend for others (RTE, NFD...) we need to document the conventions. I'm inclined to keep single-letter options with single dash (`-v`) because this is reminiscent of the unix convention.
> 
> Do we have a test which reproduces this issue?

to reproduce we can simply update the loglevel under the NROP CR and see that reflects as "--v" while "-v" is preserved as 2. 
but I see your point. what can be done then is updating this flag in the NROP side of things:
https://github.com/openshift-kni/numaresources-operator/blob/main/pkg/loglevel/loglevel.go#L57

ffromani:
> > thanks for raising this PR. Since this project is acting as frontend for others (RTE, NFD...) we need to document the conventions. I'm inclined to keep single-letter options with single dash (`-v`) because this is reminiscent of the unix convention.
> > Do we have a test which reproduces this issue?
> 
> to reproduce we can simply update the loglevel under the NROP CR and see that reflects as "--v" while "-v" is preserved as 2. but I see your point. what can be done then is updating this flag in the NROP side of things: https://github.com/openshift-kni/numaresources-operator/blob/main/pkg/loglevel/loglevel.go#L57 to set it with single "-" -> "-v"

thinking about it I may have noticed (and forgotten to follow up) something similar in the past. What's surely needed here is to document conventions and expectations (and stick to them). Afterwards we will talk code changes. There could totally be a bug in deployer, but let's clarify what should be the correct behavior before: https://github.com/k8stopologyawareschedwg/deployer/issues/254

ffromani:
this is likely a `flagcodec` bug in the end, because `pflag` comments say...
```
// PFlagFromGoFlag will return a *pflag.Flag given a *flag.Flag
// If the *flag.Flag.Name was a single character (ex: `v`) it will be accessiblei
// with both `-v` and `--v` in flags. If the golang flag was more than a single
// character (ex: `verbose`) it will only be accessible via `--verbose`
```
(from https://github.com/spf13/pflag/blob/master/golangflag.go#L60)
and this matches my experience

ffromani:
thanks for your contribution. This PR, and the related offline conversations we had, lead to https://github.com/k8stopologyawareschedwg/deployer/pull/258

PR #266, opened 2024-01-16
selinux: more policy cleanups

ffromani:
add more policy cleanups for recent OCPs, and add a command to easily inspect the proposed policy.


swatisehgal (approved):
/lgtm

PR #270, opened 2024-01-19
sched: support for informerMode option

ffromani:
The scheduler plugin gained support for the external informer, necessary to overcome  mismatch between the scheduler and the ode views wrt the pods on nodes.
Expose support in the deployer libraries to change this setting.


swatisehgal (approved):
/lgtm

PR #293, opened 2024-03-04
manifests and sched fixes

ffromani:
grab bag of fixes extracted from #291


ffromani:
better to go straight with #291

PR #294, opened 2024-03-04
DNM: WIP: TEST CI

ffromani:
DNM TEST CI


ffromani:
turns out it was a RTE issue in turn caused by vendoring a too old version of ghw

PR #311, opened 2024-08-22
OCP: SELinux: make custom policy opt-in and off by default

Tal-or:
So far, we were using a custom SELinux policy to allow RTE access to PodResourceAPI socket.

The `container_device_plugin_t` label type is a built-in context, which allows communication with
`kubelet_t` context: https://github.com/containers/container-selinux/pull/178.

The PodResourceAPI socket is an object created by Kubelet so it inherents
the same process context, i.e. `kubelet_t`.

This PR is changing RTE container to use `container_device_plugin_t` and make the logic needed for creating the custom policy as optional and off by default.

### DIFF
--- a/pkg/assets/selinux/consts.go
+++ b/pkg/assets/selinux/consts.go
@@ -18,7 +18,8 @@ package selinux
 
 const (
 	RTEPolicyFileName           = "/etc/selinux/rte.cil"
-	RTEContextType              = "rte.process"
+	RTEContextType              = "container_device_plugin_t"
+	RTEContextTypeLegacy        = "rte.process"
 	RTEContextLevel             = "s0"
 	RTEPolicyInstallServiceName = "rte-selinux-policy-install.service"
 )
--- a/pkg/manifests/rte/rte.go
+++ b/pkg/manifests/rte/rte.go
@@ -68,7 +68,10 @@ func (mf Manifests) Clone() Manifests {
 	}
 
 	if mf.plat == platform.OpenShift {
-		ret.MachineConfig = mf.MachineConfig.DeepCopy()
+		//  MachineConfig is obsolete starting from OCP v4.18
+		if mf.MachineConfig != nil {
+			ret.MachineConfig = mf.MachineConfig.DeepCopy()
+		}
 		ret.SecurityContextConstraint = mf.SecurityContextConstraint.DeepCopy()
 	}
 
@@ -110,11 +113,13 @@ func (mf Manifests) Render(opts options.UpdaterDaemon) (Manifests, error) {
 	if mf.plat == platform.OpenShift {
 		rteupdate.SecurityContext(ret.DaemonSet)
 
-		if opts.Name != "" {
-			ret.MachineConfig.Name = ocpupdate.MakeMachineConfigName(opts.Name)
-		}
-		if opts.MachineConfigPoolSelector != nil {
-			ret.MachineConfig.Labels = opts.MachineConfigPoolSelector.MatchLabels
+		if mf.MachineConfig != nil {
+			if opts.Name != "" {
+				ret.MachineConfig.Name = ocpupdate.MakeMachineConfigName(opts.Name)
+			}
+			if opts.MachineConfigPoolSelector != nil {
+				ret.MachineConfig.Labels = opts.MachineConfigPoolSelector.MatchLabels
+			}
 		}
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraint, ret.ServiceAccount)
 	}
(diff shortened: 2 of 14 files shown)


Tal-or:
Depends on: [OCPNODE-2486](https://issues.redhat.com/browse/OCPNODE-2486)

ffromani:
I'm fine fixing the policies, and there is (or we can add) logic to avoid the unconditional application of the custom policy on OCP, but I'd go for a soft-deprecation rather than a direct removal.
1. deprecate and default to off the custom policy on RTE next release
2. actually remove the support on RTE next release +1 / +2.

pkg/selinux/consts.go
```diff
@@ -17,6 +17,6 @@
 package selinux
 
 const (
-	RTEContextType  = "rte.process"
+	RTEContextType  = "container_device_plugin_t"
```
ffromani, line 20: skimming over the linked image this it seems there's a nice overlap over the features RTE needs and the features granted to this context. This seems still somehow a slight abuse (RTE is most certainly not a device plugin) but it seems the best we can do with the default contexts

ffromani:
Following the path of gradual deprecation we walked this far, we should make all these changes optional
1. MachineConfig & friends should be opt-in (because is now deprecated by default)
2. if the machine config + custom policy is disabled, the default context should be `container_device_plugin_t` and no longer `rte.process`

pkg/selinux/consts.go
Tal-or: Yes I agree, but RTE operate in a similar fashion to device plugin in the manner of communication with Kubelet, so I find this context type fair enough.

Tal-or:
> Following the path of gradual deprecation we walked this far, we should make all these changes optional
> 
> 1. MachineConfig & friends should be opt-in (because is now deprecated by default)
> 2. if the machine config + custom policy is disabled, the default context should be `container_device_plugin_t` and no longer `rte.process`

Sounds more reasonable I agree.

ffromani:
resolves: https://github.com/k8stopologyawareschedwg/deployer/issues/312

Tal-or:
@ffromani I mistakenly pushed to main (already reverted).
1. We should make main protected to direct push
2 Before we do step 1 let's remove the protection against force-push so we could delete clean the commit history from the unnecessary commits

ffromani:
> @ffromani I mistakenly pushed to main (already reverted).
> 
>     1. We should make main protected to direct push
> 
>     2. Before we do step 1 let's remove the protection against force-push so we could clean the history from the unnecessary commits

I'm against rewriting history in general.
I prefer to keep mistakes in the history (you can find plenty of mine as well). History should be immutable and append-only, because we can never know who pulled what and when.

What you did is not a big deal and a project management issue more than anything else, so it's on the maintainer (me) more than on contributor.

For these reasons I'm just enabling the branch protection without the cleanup.

pkg/manifests/rte/rte.go
```diff
@@ -68,7 +68,10 @@ func (mf Manifests) Clone() Manifests {
 	}
 
 	if mf.plat == platform.OpenShift {
-		ret.MachineConfig = mf.MachineConfig.DeepCopy()
+		//  MachineConfig is obsolete starting from 4.18v
```
ffromani, line 72: typo: v4.18

pkg/assets/selinux/consts.go
```diff
@@ -18,7 +18,8 @@ package selinux
 
 const (
 	RTEPolicyFileName           = "/etc/selinux/rte.cil"
-	RTEContextType              = "rte.process"
+	RTEContextType              = "container_device_plugin_t"
+	RTEContextObsoleteType      = "rte.process"
```
ffromani, line 22: please rename to `RTEContextTypeLegacy`

pkg/manifests/rte/rte.go
```diff
@@ -111,16 +112,18 @@ func (mf Manifests) Render(opts options.UpdaterDaemon) (Manifests, error) {
...
-
+		selinuxType := selinuxassets.RTEContextType
 		if mf.MachineConfig != nil {
 			if opts.Name != "" {
 				ret.MachineConfig.Name = ocpupdate.MakeMachineConfigName(opts.Name)
 			}
 			if opts.MachineConfigPoolSelector != nil {
 				ret.MachineConfig.Labels = opts.MachineConfigPoolSelector.MatchLabels
 			}
+			// the MachineConfig installs this custom policy which is obsolete starting from OCP 4.18v
```
ffromani, line 123: typo: v4.18

pkg/objectupdate/rte/rte.go
```diff
@@ -186,7 +186,7 @@ func SecurityContext(ds *appsv1.DaemonSet) {
 		cntSpec.SecurityContext = &corev1.SecurityContext{}
 	}
 	cntSpec.SecurityContext.SELinuxOptions = &corev1.SELinuxOptions{
-		Type:  selinuxassets.RTEContextType,
+		Type:  selinuxContextType,
```
ffromani, line 189: do we have unit test coverage? could we use a couple, and should be quick and easy to write

Tal-or: Done

ffromani (approved):
there are misplaced imports (wrong ordering), but not worth a resubmit

Tal-or:
What a bummer the /hold didn't work.
We can't use the built-in until https://issues.redhat.com/browse/OCPNODE-2486 resolved. (see: https://github.com/k8stopologyawareschedwg/deployer/pull/311#issuecomment-2303931909) 
Anyway this is not a big deal as long as we're not bumping the deployer on NROP side

ffromani:
> What a bummer the /hold didn't work. 

It can never work: we don't use [prow](https://prow.k8s.io/command-help) in this project (nor in any project in the k8stopologyawareschedwg org) because lack of resources.
We use the regular simpler github workflow here

Tal-or:
> > What a bummer the /hold didn't work.
> 
> It can never work: we don't use [prow](https://prow.k8s.io/command-help) in this project (nor in any project in the k8stopologyawareschedwg org) because lack of resources. We use the regular simpler github workflow here

Yes I remembered that the fact. I'm used to prow

PR #313, opened 2024-08-22
deployer:flags: `UpdaterCustomSELinuxPolicy`

Tal-or:
Add the `UpdaterCustomSELinuxPolicy` flag so we could control
whether we want to install custom policy from the deployer directly.

This flag is off by default.


ffromani (approved):
thanks!

ffromani:
the lack of a OCP CI lane is still annoying, but very hard to fix :\

PR #314, opened 2024-08-22
scheduler: set namespace for rolebinding

Tal-or:
The topology-aware-scheduler-leader-elect `rolebinding` object is refer to topology-aware-scheduler-leader-elect `role` which is under a specific namespace given the configuration provided to the deploy command.

we should set the right namespace for the `rolebinding` to be the same as the `role` namespace.


ffromani (approved):
makes sense, thanks for catching

PR #315, opened 2024-09-04
detect: add capability to detect HyperShift platform

Tal-or:
Added the capeability for the deployer to detect
whether the targeted OpenShift platform is of HyperShift flavor.

In future PR on NROP, we'll consume this change to allow the operator performs specific decision given the current platform.

### DIFF
--- a/pkg/deployer/platform/detect/detect.go
+++ b/pkg/deployer/platform/detect/detect.go
@@ -64,13 +64,18 @@ func Platform(ctx context.Context) (platform.Platform, error) {
 	if err != nil {
 		return platform.Unknown, err
 	}
-	return PlatformFromLister(ctx, ocpCli.ConfigV1.ClusterVersions())
+	return PlatformFromClients(ctx, ocpCli.ConfigV1.ClusterVersions(), ocpCli.ConfigV1.Infrastructures())
 }
 
 type ClusterVersionsLister interface {
 	List(ctx context.Context, opts metav1.ListOptions) (*ocpconfigv1.ClusterVersionList, error)
 }
 
+type InfrastructuresGetter interface {
+	Get(ctx context.Context, name string, options metav1.GetOptions) (*ocpconfigv1.Infrastructure, error)
+}
+
+// PlatformFromLister is deprecated, use PlatformFromClients instead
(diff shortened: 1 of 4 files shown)


pkg/deployer/platform/detect/detect.go
```diff
@@ -64,14 +64,18 @@ func Platform(ctx context.Context) (platform.Platform, error) {
...
 type ClusterVersionsLister interface {
 	List(ctx context.Context, opts metav1.ListOptions) (*ocpconfigv1.ClusterVersionList, error)
 }
 
-func PlatformFromLister(ctx context.Context, cvLister ClusterVersionsLister) (platform.Platform, error) {
+type InfrastructuresGetter interface {
+	Get(ctx context.Context, name string, options metav1.GetOptions) (*ocpconfigv1.Infrastructure, error)
+}
+
+func PlatformFromClients(ctx context.Context, cvLister ClusterVersionsLister, infraGetter InfrastructuresGetter) (platform.Platform, error) {
```
ffromani, line 78: let's try not to break backward compatibility so suddenly. Let's keep the old function (which will never return Hypershift) and let's add this one. When you do so please add a comment on `PlatformFromLister` like "deprecated, use PlatformFromClients", so in the next-next release or later we can remove it

ffromani (commented):
LGTM except for breaking backward compat. Let's handle it and we can merge.

pkg/deployer/platform/detect/detect.go
Tal-or: Ack. what about the unit-tests? keep the old one?

ffromani: > Ack. what about the unit-tests? keep the old ones as well?

every public function should be unit-tested if possible, unless there are trivial wrappers kept for backward compat. So I'd say yes, let's keep (some) old unit tests

Tal-or:
> /hold

Wanted to check something the diff seemed suspicious. PR is reviewable now.

ffromani (approved):
detecting openshift sub-families is useful

ffromani:
I'd be open to a backport to the `release-0.20` branch

PR #327, opened 2024-11-20
rte: avoid \`MachineConfig\` manifest for HyperShift

Tal-or:
There's a bug in `GetManifest` that it might
add `MachineConfig` manifest while running on HyperShift platform.

This case is always invalid because HyperShift is not familiar with the `MachineConfig` term at all.

### DIFF
--- a/pkg/manifests/rte/rte.go
+++ b/pkg/manifests/rte/rte.go
@@ -185,14 +185,13 @@ func GetManifests(plat platform.Platform, version platform.Version, namespace st
 	var err error
 	mf := New(plat)
 
-	if plat == platform.OpenShift || plat == platform.HyperShift {
-		if withCustomSELinuxPolicy {
-			mf.MachineConfig, err = manifests.MachineConfig(manifests.ComponentResourceTopologyExporter, version, withCRIHooks)
-			if err != nil {
-				return mf, err
-			}
+	if plat == platform.OpenShift && withCustomSELinuxPolicy {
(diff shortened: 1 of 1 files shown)


ffromani (approved):
thanks!

pkg/manifests/rte/rte.go
```diff
@@ -185,14 +185,13 @@ func GetManifests(plat platform.Platform, version platform.Version, namespace st
 	var err error
 	mf := New(plat)
 
-	if plat == platform.OpenShift || plat == platform.HyperShift {
-		if withCustomSELinuxPolicy {
-			mf.MachineConfig, err = manifests.MachineConfig(manifests.ComponentResourceTopologyExporter, version, withCRIHooks)
-			if err != nil {
-				return mf, err
-			}
+	if plat == platform.OpenShift && withCustomSELinuxPolicy {
```
shajmakh, line 188: do we prefer t he condition to be `plat == platform.OpenShift` or `plat != platform.Hypershift` ?
I'm asking because it was mentioned earlier that platform.Kubernetes is allows with MCP installed so I wonder if we want to still consider this case

ffromani: Good point. The best option is to review all the platform checks and make sure we are consistent. This change seems to follow the spirit of the old code so should be ok.

Tal-or: This MachineConfig intended to install the custom SELinux policy which tied up to the RHEL OS which is running on the system. 
This is OCP specific

PR #331, opened 2024-12-09
selinux: add kubelet_var_lib_t permissions

Tal-or:
In future rhel/rhcos release podresourceAPI socket file's context might change to `kubelet_var_lib_t`.

To deal with this issue in advance we'll change RTE custom policy to support this new context as well.


ffromani (approved):
thanks!

PR #320, opened 2024-09-24
selinux: allow access to kubelet_var_lib_t

Tal-or:
podresources API context has changed to kubelet_var_lib_t:
 https://github.com/containers/container-selinux/pull/329

Adjust the custom policy to allow access to this new file context.

Signed-off-by: Talor Itzhak <titzhak@redhat.com>

PR #341, opened 2024-12-16
[release-0.20][manual] support for custom and default selinux policies

ffromani:
Aggregate backport of #311 #313 #319 (partial) #320 (partial)


Tal-or:
LGTM
let's merge after #344 goes in

ffromani:
let's see if we can avoid the backports because it creates almost more problems as it solves it

PR #350, opened 2025-02-17
SCC: Added annotation to daemonset pods template

rbaturov:
Since OCP 4.18, all workloads of platform namespaces (openshift-*) explicitly pin the SCCs they need.

https://docs.openshift.com/container-platform/4.17/authentication/managing-security-context-constraints.html#security-context-constraints-requiring_configuring-internal-oauth

https://github.com/openshift/origin/blob/master/pkg/monitortests/authentication/requiredsccmonitortests/monitortest.go

### DIFF
--- a/pkg/manifests/rte/rte.go
+++ b/pkg/manifests/rte/rte.go
@@ -125,7 +125,13 @@ func (mf Manifests) Render(opts options.UpdaterDaemon) (Manifests, error) {
 		}
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraint, ret.ServiceAccount)
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraintV2, ret.ServiceAccount)
-		rteupdate.SecurityContext(ret.DaemonSet, selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)))
+		rteupdate.SecurityContextWithOpts(
+			ret.DaemonSet,
+			rteupdate.SecurityContextOptions{
+				SELinuxContextType:  selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)),
+				SecurityContextName: mf.SecurityContextConstraint.Name,
+			},
+		)
 	}
 
 	return ret, nil
--- a/pkg/objectupdate/rte/rte.go
+++ b/pkg/objectupdate/rte/rte.go
@@ -46,6 +46,7 @@ const (
 	rteKubeletDirVolumeName      = "host-var-lib-kubelet"
 	rteNotifierFileName          = "notify"
 	hostNotifierDir              = "/run/rte"
+	sccAnnotation                = "openshift.io/required-scc"
 )
 
 func ContainerConfig(podSpec *corev1.PodSpec, cnt *corev1.Container, configMapName string) {
@@ -188,6 +189,37 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
 	cntSpec.SecurityContext.SELinuxOptions = &corev1.SELinuxOptions{
 		Type:  selinuxContextType,
 		Level: selinuxassets.RTEContextLevel,
+	}
+}
+
+type SecurityContextOptions struct {
+	SELinuxContextType  string
+	SecurityContextName string
+}
+
+func SecurityContextWithOpts(ds *appsv1.DaemonSet, opts SecurityContextOptions) {
+	cntSpec := objectupdate.FindContainerByName(ds.Spec.Template.Spec.Containers, manifests.ContainerNameRTE)
+	if cntSpec == nil {
+		return
+	}
+
+	// this is needed to put watches in the kubelet state dirs AND
+	// to open the podresources socket in R/W mode
+	if cntSpec.SecurityContext == nil {
+		cntSpec.SecurityContext = &corev1.SecurityContext{}
+	}
+	if opts.SELinuxContextType != "" {
+		cntSpec.SecurityContext.SELinuxOptions = &corev1.SELinuxOptions{
+			Type:  opts.SELinuxContextType,
+			Level: selinuxassets.RTEContextLevel,
+		}
+	}
+
+	if ds.Spec.Template.ObjectMeta.Annotations == nil {
+		ds.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+	if opts.SecurityContextName != "" {
+		ds.Spec.Template.ObjectMeta.Annotations[sccAnnotation] = opts.SecurityContextName
 	}
 }
 
--- a/pkg/objectupdate/rte/rte_test.go
+++ b/pkg/objectupdate/rte/rte_test.go
@@ -18,14 +18,14 @@ package rte
 
 import (
 	"fmt"
-	selinuxassets "github.com/k8stopologyawareschedwg/deployer/pkg/assets/selinux"
-	"github.com/k8stopologyawareschedwg/deployer/pkg/objectupdate"
 	"strings"
 	"testing"
 	"time"
 
+	selinuxassets "github.com/k8stopologyawareschedwg/deployer/pkg/assets/selinux"
 	"github.com/k8stopologyawareschedwg/deployer/pkg/deployer/platform"
 	"github.com/k8stopologyawareschedwg/deployer/pkg/manifests"
+	"github.com/k8stopologyawareschedwg/deployer/pkg/objectupdate"
 	"github.com/k8stopologyawareschedwg/deployer/pkg/options"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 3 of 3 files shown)


ffromani (commented):
thanks for the change! Let's make sure it is covered by unit tests (existing or new)

pkg/objectupdate/rte/rte.go
```diff
@@ -150,6 +150,11 @@ func DaemonSet(ds *appsv1.DaemonSet, plat platform.Platform, configMapName strin
 		podSpec.NodeSelector = opts.NodeSelector.MatchLabels
 	}
 	MetricsPort(ds, metricsPort)
+
+	if ds.Spec.Template.Annotations == nil {
+		ds.Spec.Template.Annotations = make(map[string]string)
+	}
+	ds.Annotations["openshift.io/required-scc"] = "resource-topology-exporter"
```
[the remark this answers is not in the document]
rbaturov: Actually we can take it from:
`DaemonSet.Spec.Template.Spec.ServiceAccountName` which suppose to be identical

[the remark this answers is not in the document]
rbaturov: OR `mf.ServiceAccount.Name`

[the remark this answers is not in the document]
rbaturov: Yes we can take it from the `securitycontextconstraint.yaml`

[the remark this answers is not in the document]
rbaturov: We can do it [here](https://github.com/rbaturov/deployer/blob/add-scc-annotation/pkg/manifests/rte/rte.go#L104) the same way

pkg/manifests/rte/rte.go
```diff
@@ -126,6 +126,11 @@ func (mf Manifests) Render(opts options.UpdaterDaemon) (Manifests, error) {
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraint, ret.ServiceAccount)
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraintV2, ret.ServiceAccount)
 		rteupdate.SecurityContext(ret.DaemonSet, selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)))
+
+		if ret.DaemonSet.Spec.Template.ObjectMeta.Annotations == nil {
+			ret.DaemonSet.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+		}
+		ret.DaemonSet.Spec.Template.ObjectMeta.Annotations["openshift.io/required-scc"] = mf.SecurityContextConstraint.Name
```
ffromani, line 133: this should go inside `rteupdate.DaemonSet`

rbaturov: To do this I'd need to pass the manifests object as an argument to that function. Therefore we would need to import the `manifests/rte` which causes import cycle.
We can also pass the name of the scc directly to the `rteupdate.DaemonSet`, but this is pretty odd.

ffromani: passing the manifest object is not a good idea indeed. We can add however a `SCCName` field in the `options.DaemonSet` struct and use that to make logic.
I'm not sure I get why passing the SCC Name would be odd

rbaturov: Adding it to the `options.DaemonSet`, will require to add it [here](https://github.com/rbaturov/deployer/blob/main/pkg/options/options.go#L130) 
 aswell. This will require setting it up in the `commonOpts` struct.
 As far as I understand we populate this struct from the command line flags.

Maybe we should just add the SCC name as an argument to `rteupdate.DaemonSet`

ffromani: uhm, I'm now thinking if this doesn't belong more to `rteupdate.SecurityContext`.
I'm also pondering if we should have it moved in `ocpupdate`. But I'm inclined about just extending `rteupdate.SecurityContext`. We should try to maximize backward compat, so we should preobably add another function taking a struct with parameters (`name` and `type` probably) and deprecate the current `SecurityContext`.

rbaturov: Therefore, creating this new function will require us to bump the deployer on NROP and change [these calls](https://github.com/rbaturov/numaresources-operator/blob/main/internal/controller/numaresourcesoperator_controller.go#L749-L752), to call the new function instead. Did I got this correctly?

ffromani: yes, the gist is correct. Adding new functions is almost always fine.

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +191,28 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
 	}
 }
 
+func SecurityContextWithSCC(ds *appsv1.DaemonSet, selinuxContextType, sccName string) {
```
ffromani, line 194: I feel using the options pattern (see: https://dev.to/kittipat1413/understanding-the-options-pattern-in-go-390c et. al.) is too much here.
But we can and should add an `SCCOptions` struct and use it in the extended function:
```
type SecurityContextOptions struct {
    SELinuxContextType string
    SecurityContextName string
}

func SecurityContextWithOpts(ds *appsv1.DaemonSet, opts SecurityContextOptions) { 
...
}
```

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +191,28 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
...
+	}
+	cntSpec.SecurityContext.SELinuxOptions = &corev1.SELinuxOptions{
+		Type:  selinuxContextType,
+		Level: selinuxassets.RTEContextLevel,
+	}
+
+	if ds.Spec.Template.ObjectMeta.Annotations == nil {
+		ds.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+	ds.Spec.Template.ObjectMeta.Annotations["openshift.io/required-scc"] = sccName
```
ffromani, line 213: `"openshift.io/required-scc"` should be a private constant in this file

pkg/objectupdate/rte/rte.go
rbaturov:
```
type SecurityContextOptions struct {
    SELinuxContextType string
    SecurityContextName string
}
```

Should we place this struct in this file?
I wonder if we have a better place

rbaturov:
@ffromani Thanks for the review. Code updated.

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +192,33 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
 	}
 }
 
+type SecurityContextOptions struct {
+	SELinuxContextType  string
+	SecurityContextName string
+}
+
+func SecurityContextWithOpts(ds *appsv1.DaemonSet, opts *SecurityContextOptions) {
```
ffromani, line 200: should be passed per value not per pointer

pkg/manifests/rte/rte.go
```diff
@@ -125,7 +125,8 @@ func (mf Manifests) Render(opts options.UpdaterDaemon) (Manifests, error) {
 		}
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraint, ret.ServiceAccount)
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraintV2, ret.ServiceAccount)
-		rteupdate.SecurityContext(ret.DaemonSet, selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)))
+		sccOpts := &rteupdate.SecurityContextOptions{SELinuxContextType: selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)), SecurityContextName: mf.SecurityContextConstraint.Name}
+		rteupdate.SecurityContextWithOpts(ret.DaemonSet, sccOpts)
```
ffromani, line 129: you can inline:
```
rteupdate.SecurityContextWithOpts(ret.DaemonSet, rteupdate.SecurityContextOptions{
    SELinuxContextType: selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)), 
    SecurityContextName: mf.SecurityContextConstraint.Name,
})
```

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +192,33 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
 	}
 }
 
+type SecurityContextOptions struct {
```
ffromani, line 195: placement is fine right here

pkg/objectupdate/rte/rte.go
rbaturov: What is the reason? I want to learn why

rbaturov: Is this to support your inline suggestion down bellow? or some principle I'm not following

ffromani: because when functions accept a pointer argument, the general implied meaning is "the function can mutate this argument", which is not the intention nor the goal here. The function will never mutate the options.
Only if and when the struct passed by argument became really huge (but why did we let that  happen?) we can evaluate to pass it per pointer only for performance reasons.

Everything else equal, the readability requirement trumps any performance requirement, especially about preemptive (= not backed by profiling data) optimizations.

pkg/manifests/rte/rte.go
```diff
@@ -125,7 +125,7 @@ func (mf Manifests) Render(opts options.UpdaterDaemon) (Manifests, error) {
 		}
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraint, ret.ServiceAccount)
 		ocpupdate.SecurityContextConstraint(ret.SecurityContextConstraintV2, ret.ServiceAccount)
-		rteupdate.SecurityContext(ret.DaemonSet, selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)))
+		rteupdate.SecurityContextWithOpts(ret.DaemonSet, rteupdate.SecurityContextOptions{SELinuxContextType: selinuxTypeFromSCCVersion(opts.DaemonSet.SCCVersion, (mf.MachineConfig != nil)), SecurityContextName: mf.SecurityContextConstraint.Name})
```
ffromani, line 128: any reason to inline the code like this vs the expanded canonical struct literal definition?

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +192,33 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
...
+	}
+	cntSpec.SecurityContext.SELinuxOptions = &corev1.SELinuxOptions{
+		Type:  opts.SELinuxContextType,
+		Level: selinuxassets.RTEContextLevel,
+	}
+
+	if ds.Spec.Template.ObjectMeta.Annotations == nil {
+		ds.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
+	ds.Spec.Template.ObjectMeta.Annotations[sccAnnotation] = opts.SecurityContextName
```
ffromani, line 219: this block should be done only `if opts.SecurityContextName != ""`

ffromani: please add a test to check that if the name is `""` no annotations is added

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +192,33 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
...
+		return
+	}
+
+	// this is needed to put watches in the kubelet state dirs AND
+	// to open the podresources socket in R/W mode
+	if cntSpec.SecurityContext == nil {
+		cntSpec.SecurityContext = &corev1.SecurityContext{}
+	}
+	cntSpec.SecurityContext.SELinuxOptions = &corev1.SELinuxOptions{
+		Type:  opts.SELinuxContextType,
```
ffromani, line 212: this must be `!= ""` and the function should error out if it is nil (yes, this is stricter than the original function)

pkg/manifests/rte/rte.go
rbaturov: Not really, updated this.

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +192,37 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
...
+	cntSpec := objectupdate.FindContainerByName(ds.Spec.Template.Spec.Containers, manifests.ContainerNameRTE)
+	if cntSpec == nil {
+		return
+	}
+
+	// this is needed to put watches in the kubelet state dirs AND
+	// to open the podresources socket in R/W mode
+	if cntSpec.SecurityContext == nil {
+		cntSpec.SecurityContext = &corev1.SecurityContext{}
+	}
```
ffromani, line 210: we should not mutate the `ds` unnecessarily, but we can fix this later.

pkg/objectupdate/rte/rte.go
```diff
@@ -191,6 +192,37 @@ func SecurityContext(ds *appsv1.DaemonSet, selinuxContextType string) {
...
+	if opts.SELinuxContextType != "" {
+		cntSpec.SecurityContext.SELinuxOptions = &corev1.SELinuxOptions{
+			Type:  opts.SELinuxContextType,
+			Level: selinuxassets.RTEContextLevel,
+		}
+	}
+
+	if ds.Spec.Template.ObjectMeta.Annotations == nil {
+		ds.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+	}
```
ffromani, line 220: we should not mutate the `ds` unnecessarily, but we can fix this later.

pkg/objectupdate/rte/rte_test.go
```diff
@@ -349,3 +349,63 @@ func TestSecurityContext(t *testing.T) {
 		})
 	}
 }
+
+func TestSecurityContextWithOpts(t *testing.T) {
+	testCases := []struct {
+		description        string
+		selinuxContextType string
+		sccName            string
+	}{
```
ffromani, line 358: would have been better to have a reference object to compare equality for in the test body, but also this approach is acceptable.

ffromani (approved):

PR #357, opened 2025-05-15
rte: selinux: missing v2 SCC creation and deletion

rbaturov:
This PR complements the work done in #348.
The sections added were missing, which subsequently prevented the SCC v2 from being deployed.

### DIFF
--- a/pkg/objectwait/rte/rte.go
+++ b/pkg/objectwait/rte/rte.go
@@ -42,6 +42,12 @@ func Creatable(mf rtemf.Manifests, cli client.Client, log logr.Logger) []objectw
 		})
 	}
 
+	if mf.SecurityContextConstraintV2 != nil {
+		objs = append(objs, objectwait.WaitableObject{
+			Obj: mf.SecurityContextConstraintV2,
+		})
+	}
+
 	if mf.MachineConfig != nil {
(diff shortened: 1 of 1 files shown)


rbaturov:
@ffromani @swatisehgal @Tal-or Can/Should we add some e2e coverage here?

P.S. I wonder if we could leverage AI to identify these issues or help highlight all the relevant parts of the code that might need updating, especially since this isn’t the first time it’s happened.

ffromani (commented):
yes, this seems an omission of the original PR

ffromani:
> P.S. I wonder if we could leverage AI to identify these issues or help highlight all the relevant parts of the code that might need updating, especially since this isn’t the first time it’s happened.

I think that the gaps are very well known here: lack of OCP CI at all and underpowered kube CI

Tal-or:
I don't remember we we already have it but running a render flow and check all manifests have been created should be sufficient enough to catch error like this one

ffromani:
another option: reimplement render reusing the create flow. This at least enables us to test it. Usinf a fake client may be the simplest way.

ffromani (approved):

ffromani:
@rbaturov https://github.com/k8stopologyawareschedwg/deployer/issues/358

PR #361, opened 2025-07-06
manifests: add network policies (UPDATED)

Tal-or:
Picking up the work from https://github.com/k8stopologyawareschedwg/deployer/pull/355


Tal-or:
The existing kind network plugin doesn't seems to support network policies enforcement.
I did the following steps to prove my theory:

Created a `NetworkPolicy` to block all egress traffic for pod on `default` namespace
```
cat << EOF | kubectl create -f -                                                                                                                                      
# deny-egress.yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: deny-all-egress
  namespace: default
spec:                                                                                                                                                                                         
  podSelector: {}                                                                                                                                                                             
  policyTypes:
  - Egress                                                                                                                                                                                    
              
EOF
```

Created simple pod that has curl
```
cat << EOF | kubectl create -f -
# pod.yaml
apiVersion: v1
kind: Pod
metadata:
  name: curlpod
  labels:
    app: curl
  namespace: default
spec:
  containers:
  - name: curl
    image: curlimages/curl:7.88.1
    command: ["sleep", "3600"]

EOF
```
Access the pod
```
kubectl exec -it curlpod -- sh
```

Run curl to google successfully:
```
/ $ curl https://google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
</BODY></HTML>
```

So even when we have network policy apply it's not supported by kind network plugin.

PR #375, opened 2025-09-17
selinux: allow access to kubelet_var_lib_t in OCP 4.17

shajmakh:
In OCP 4.17 the container-selinux version was bumped to v2.237 which uses kubelet_var_lib_t instead of container_var_lib_t in older versions. Add a new access allowance for RTE to this socket file to enable RTE functionality on the newer builds of 4.17.


shajmakh:
wait for https://github.com/openshift-kni/numaresources-operator/pull/2067 verification

